PR #4, opened 2024-03-08
Promote vmware guest

machacekondra:


mariolenz:
@machacekondra Are you _really_ sure you want to have `vmware_guest` in this collection? As it looks now? Not to put too fine a point on this but IMHO the module tries to do too much, is far too complex (I always think about it as a [god module](https://en.wikipedia.org/wiki/God_object)) and not maintainable. People use and like it the way it is, but I think it's pretty problematic.

If you really want to have it, I'm fine with this. If there's one module I'd like to get rid of, it's this one. For the reasons I've stated above. But think twice about it. It's your funeral ;-)

machacekondra:
> @machacekondra Are you _really_ sure you want to have `vmware_guest` in this collection? As it looks now? Not to put too fine a point on this but IMHO the module tries to do too much, is far too complex (I always think about it as a [god module](https://en.wikipedia.org/wiki/God_object)) and not maintainable. People use and like it the way it is, but I think it's pretty problematic.
> 
> If you really want to have it, I'm fine with this. If there's one module I'd like to get rid of, it's this one. For the reasons I've stated above. But think twice about it. It's your funeral ;-)

I beleive we want to have it, but of course I agree it needs refactoring. How would you propose to make this module more maintanable? Create multiple modules out of it, or just refactor the code and keep the functinality? I beleive our end goal is to keep the interface of the modules, so users just rename the namespace and keep the rest of parameters as is.

mariolenz:
> I beleive we want to have it

Yes, I thought so. I don't have any data, but I guess it's a module that a lot of people use. Maybe even the _most used_ module... but this is just a guess.

> but of course I agree it needs refactoring. How would you propose to make this module more maintanable? Create multiple modules out of it, or just refactor the code and keep the functinality?

If I had an idea how to fix this, I'd have already started on working on implementing it in c.vmware :wink: 

Some ideas:

1. Drop support for adding _a lot_ of devices when creating a VM. For example, disks. Make people create a VM and use `vmware_guest_controller` and `vmware_guest_disk` in following steps.
2. Drop some other configuration an make people use other modules to achieve this. For example, drop `boot_firmware` and `secure_boot`; people can use `vmware_guest_boot_manager` in a second step to configure this.
3. Move cloning a VM / template to another module, or maybe two other modules.

It would be a lot of work to figure out what to drop and what to keep, and I'm pretty sure those changes would be extremely unpopular. But I think it would be the only way forward.

> I beleive our end goal is to keep the interface of the modules, so users just rename the namespace and keep the rest of parameters as is, so we keep the migration path for them as simple as possible.

I don't think you can have both: A module that's somewhat easy to maintain and keeping it as-is. Refactoring won't help you there. The module just tries to do too many things at once. My personal opinion: Drop functionality and make users angry, or keep functionality and live with a module far too complex. I might be wrong, but at the moment that's how I see things.

machacekondra:
> If I had an idea how to fix this, I'd have already started on working on implementing it in c.vmware 😉
> 
> Some ideas:
> 
>     1. Drop support for adding _a lot_ of devices when creating a VM. For example, disks. Make people create a VM and use `vmware_guest_controller` and `vmware_guest_disk` in following steps.
> 
>     2. Drop some other configuration an make people use other modules to achieve this. For example, drop `boot_firmware` and `secure_boot`; people can use `vmware_guest_boot_manager` in a second step to configure this.
> 
>     3. Move cloning a VM / template to another module, or maybe two other modules.
> 
> 
> It would be a lot of work to figure out what to drop and what to keep, and I'm pretty sure those changes would be extremely unpopular. But I think it would be the only way forward.

So what about keeping the `vmware_guest` in a `c.vmware` and creating better modules in `vmware.vmware`? 
We also have created follwoing collection:
 https://github.com/redhat-cop/cloud.vmware_ops
 
There is a role called `provison_vm`, which basically just calls the `c.vmware_guest` module. So the interface of the role is very similar to interface of `c.vmware_guest`. With that we can in future change the implementation of the role to use new modules in `vmware.vmware` without changing the interface and keep the similar functionality as `c.vmware_guest`. Maybe with such an approach we won't upset too many users.

mariolenz:
> So what about keeping the `vmware_guest` in a `c.vmware` and creating better modules in `vmware.vmware`?

At the moment, I guess this would be the best idea. Although I'm not sure if you really have to create (all) those modules. I guess there are already some in c.vmware that we could move. I'll try to find the time to have a closer look at this, but probably not before the weekend.

Of course, this would break the

> goal to keep the interface of the modules, so users just rename the namespace and keep the rest of parameters as is, so we keep the migration path for them as simple as possible

Maybe we should discuss this further when I've found the time to make some proposals. But, as I've said, I don't think I'll find the time to do this before the weekend.

mariolenz:
This module is so complex that even thinking about how to improve it hurts...

OK, I should say that the main problem is that c.vmware_guest tries to do too many things at once. It not only makes sure a VM exists / doesn't exist, it also manages some basic configuration like CPUs and memory (which should be OK), can mark a VM as a template and vice versa, manages CDROMs / disks / NICs along with advanced settings / annotations / custom attributes and some other stuff.

Maybe we should strip this down considerably to creating / deleting VMs and managing some very basic configuration. I've started to play around with this idea, but it's still early days: ansible-collections/community.vmware#2052

mariolenz:
There's both [select_datastore](https://github.com/ansible-collections/community.vmware/blob/e0bd265f2902aad46cd2cabf17a786e074be0488/plugins/modules/vmware_guest.py#L2738) and [autoselect_datastore](https://github.com/ansible-collections/community.vmware/blob/e0bd265f2902aad46cd2cabf17a786e074be0488/plugins/modules/vmware_guest.py#L2720). I don't really understand this ATM.

As I've said, this module is a mess :shrug:

mariolenz:
Argh! The documentation says that you need to define either `esxi_hostname` or `cluster`, but there are even _CI tests_ that don't and still work. I'm afraid changing this would be a breaking change :cry: 

[Burn it!](https://youtu.be/wYHWF5NjO1k)

PR #10, opened 2024-04-16
Fix CI unit tests

machacekondra:
.. and remove duplicated workflow to run unit tests and integration tests


.github/workflows/ansible-test.yml
```diff
@@ -110,148 +110,13 @@ jobs:
           # https://github.com/ansible-community/ansible-test-gh-action#pull-request-change-detection
           pull-request-change-detection: false
 
-###
-# Unit tests (OPTIONAL)
```
bardielle, line 114: Does it mean that you removed the UT? there is not UT running at all after that PR?

machacekondra: Well, I removed duplicated UT, there are already UT running in `github/workflows/test.yml`

bardielle: Maybe we can change the file names? 
ansible-test and test are confusing names... maybe ansible-sanity.yml?

machacekondra: Yeah, that would be better.

PR #12, opened 2024-04-21
Copy dynamic inventory

bardielle:
##### SUMMARY
Copy dynamic inventory from community.vmware

### DIFF
--- a/scripts/inventory/vmware.ini
+++ b/scripts/inventory/vmware.ini
@@ -1,0 +1,50 @@
+# Ansible VMware external inventory script settings
+
+[defaults]
+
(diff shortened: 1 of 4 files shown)


mariolenz:
I'm afraid that this is a part of the community.vmware collection that I've never really understood. If you think this is a problem because I'm the maintainer of the collection, I agree. But I want to put my cards on the table (I'm not sure if this correct, English isn't my first language; in German, it's literally "playing with open cards" meaning that I don't want to hide anything).

Anyway, I don't think I can help you there. Possibly with everything else, but this inventory stuff is my blind spot :shrug:

bardielle:
@mariolenz thank you for your feedback and the honesty.

PR #15, opened 2024-05-12
Require pyVmomi 6.7.1

mariolenz:
##### SUMMARY
Some modules in [community.vmware](https://github.com/ansible-collections/community.vmware) require pyVmomi 6.7.1 or later. I think it would be easier to move those modules in the future to this collection (if we want to) if we already define this in `requirements.txt` now.

Additionally, `HAS_PYVMOMIJSON` is only used to test this and we could get rid of this because a) it's not used anywhere yet and b) wouldn't be needed if we define `pyVmomi>=6.7.1` in `requirements.txt`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
requirements.txt

##### ADDITIONAL INFORMATION
ansible-collections/community.vmware#2071

[6.7.1](https://pypi.org/project/pyvmomi/6.7.1/) has been released 2018-10-23, 5 1/2 years ago.

### DIFF
--- a/changelogs/fragments/15-clarify-pyvmomi-requirement.yml
+++ b/changelogs/fragments/15-clarify-pyvmomi-requirement.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - Clarify pyVmomi requirement (https://github.com/ansible-collections/vmware.vmware/pull/15).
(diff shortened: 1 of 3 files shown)


machacekondra (approved):
Thanks Mario!

PR #16, opened 2024-05-15
Add vm_list_group_by_clusters module

bardielle:
##### SUMMARY
New module vm_list_group_by_clusters

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
vm_list_group_by_clusters

##### ADDITIONAL INFORMATION
This module is building a map of clusters, in each cluster there is a map of folders that includes list of the VMs. 

This module bring a list of all the VMs for MTV process.


plugins/modules/vm_list_group_by_clusters.py
```diff
@@ -0,0 +1,191 @@
...
+short_description: Returns information about the virtual machines grouping by clusters and folders
+description:
+    - Returns information about the virtual machines grouping by clusters and folders.
+author:
+    - Ansible Cloud Team (@ansible-collections)
+options:
+  detailed_vms:
+    default: true
+    description:
+      - Wither or not get the vms detailed.
```
machacekondra, line 24: I would write something like: If I(true) gather detailed information about virtual machines.

plugins/modules/vm_list_group_by_clusters.py
```diff
@@ -0,0 +1,191 @@
...
+DOCUMENTATION = r'''
+---
+module: vm_list_group_by_clusters
+short_description: Returns information about the virtual machines grouping by clusters and folders
+description:
+    - Returns information about the virtual machines grouping by clusters and folders.
+author:
+    - Ansible Cloud Team (@ansible-collections)
+options:
+  detailed_vms:
```
machacekondra, line 21: Please keep the spacing the same. Above you use 4 spaces, and for example below just 2 spaces.

plugins/modules/vm_list_group_by_clusters.py
```diff
@@ -0,0 +1,191 @@
...
+'''
+
+EXAMPLES = r'''
+- name: Gather list of VMs
+  vmware.vmware.vm_list_group_by_clusters:
+    hostname: "https://vcenter"
+    username: "username"
+    password: "password"
+
+- name: Gather list of VMs in clusterA
```
machacekondra, line 38: I think this description is wrong. The module don't support grouping by specific cluster.

plugins/modules/vm_list_group_by_clusters.py
```diff
@@ -0,0 +1,191 @@
...
+                    if self.detailed_vms:
+                        vm = self.get_vm_detailed(vm=vm.vm)
+
+                    vms_detailed.append(self._vvars(vm))
+
+                if vms_detailed:
+                    empty = False
+                    vm_list_group_by_folder_dict[folder.name] = vms_detailed
+
+            if not empty:
```
machacekondra, line 164: If I understand correctly you can get rid of the `empty` variable, and just check if `vm_list_group_by_folder_dict` is empty.

tests/integration/targets/vm_list_group_by_clusters/mock.json
```diff
@@ -0,0 +1,106 @@
...
+  "priority": 1
+},
+{
+  "httpRequest": {
+    "method": "POST",
+    "path": "/api",
+    "body": {
+      "type": "JSON",
+      "matchType": "PARTIAL",
+      "json": "{\"params\":{\"serviceId\":\"com.vmware.cis.session\",\"operationId\":\"get\"}}"
```
machacekondra, line 80: Is this call really called? I dind't see it before. Anyway, I think the result output is wrong, as it's copy paste of the hostname call.

tests/integration/targets/vm_list_group_by_clusters/mock.json
```diff
@@ -0,0 +1,106 @@
...
+    "method": "POST",
+    "path": "/api",
+    "body": {
+      "type": "JSON",
+      "matchType": "PARTIAL",
+      "json": "{\"params\":{\"serviceId\":\"com.vmware.vcenter.VM\",\"operationId\":\"list\"}}"
+    }
+  },
+  "httpResponseTemplate": {
+    "template": "{\"statusCode\": 200, \"headers\": {\"Content-type\": \"application/json\"}, \"body\": {\"jsonrpc\": \"2.0\", \"result\": {\"output\": [{\"STRUCTURE\": {\"com.vmware.vcenter.VM\": {\"name\": \"vm1\", \"power_state\": \"POWERED_OFF\", \"vm\": \"vm1\"}}}]}, 'id': '2'}}",
```
machacekondra, line 68: Please use `${json.parse($!request.body)['id']}'` for `id`, it may work now with hardcoded numbers but, may be flaky in future. same for above calls.

tests/integration/targets/vm_list_group_by_clusters/vars.yml
```diff
@@ -0,0 +1,6 @@
+vcenter_hostname: "127.0.0.1"
+vcenter_username: "user"
+vcenter_password: "pass"
+vcenter_port: 1080
+
+mock_file: "vm_list_group_by_clusters"
```
machacekondra, line 6: Please add new line.

tests/integration/targets/vm_list_group_by_clusters/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+    password: "{{ vcenter_password }}"
+    detailed_vms: false
+    port: "{{ vcenter_port }}"
+  register: __res
+
+- name: Assert values
+  ansible.builtin.assert:
+    that:
+      - __res.changed == False
+      - __res.vm_list_group_by_clusters | length == 1
```
machacekondra, line 16: Can you please also add:
 `- __res.vm_list_group_by_clusters['cluster1'] | length == 1` 
 `- __res.vm_list_group_by_clusters['cluster1']['folder1'] | length == 1`

plugins/modules/vm_list_group_by_clusters.py
bardielle: removed that example

tests/integration/targets/vm_list_group_by_clusters/mock.json
bardielle: It wasn't work for me... that's why I moved to use the ID

tests/integration/targets/vm_list_group_by_clusters/mock.json
bardielle: removed it

machacekondra (approved):

PR #18, opened 2024-05-16
Run CI with core 2.17

mariolenz:
I think for the inclusion into the community package / Ansible we will be asked to test with ansible-core 2.17. Let's implement this.


machacekondra (approved):
Thank you!

PR #19, opened 2024-05-17
Add vcsa_settings module

machacekondra:
System module is used to Configure system settings of VMware appliance.

### DIFF
--- a/plugins/modules/vcsa_settings.py
+++ b/plugins/modules/vcsa_settings.py
@@ -1,0 +1,576 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2023, Ansible Cloud Team (@ansible-collections)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: vcsa_settings
+short_description: Configure vCenter Server Appliance settings
+version_added: 2.0.0
+description:
+  - Configure vCenter Server Appliance settings.
+author:
+  - Ansible Cloud Team (@ansible-collections)
+options:
+  timezone:
+    description:
+      - Set time zone.
+    type: str
+  global_fips:
+    description:
+      - "Enable/Disable Global FIPS mode for the appliance. Caution: Changing the value of this setting will reboot the Appliance."
+    type: bool
+  resize_storage:
+    description:
+      - Resize all partitions to 100 percent of disk size.
+    type: bool
+    default: false
+  dns_mode:
+    choices:
+      - is_static
+      - dhcp
+    description:
+      - Set the DNS mode.
+    type: str
+  dns_append:
+    description:
+      - If I(true) items from C(dns_domains) and C(dns_servers) will be added to already configured DNS domains/servers.
+      - If I(false) domains/servers will be overriden.
+    type: bool
+    default: true
+  dns_hostname:
+    description:
+      - DNS hostname.
+    type: str
+  dns_servers:
+    description:
+      - List of DNS servers.
+    type: list
+    elements: str
+  dns_domains:
+    description:
+      - List of DNS domains.
+    type: list
+    elements: str
+  timesync_mode:
+    description:
+      - Set time synchronization mode.
+    type: str
+    choices:
+      - disabled
+      - host
+      - ntp
+  ntp_servers:
+    description:
+      - List of NTP servers. This method updates old NTP servers from configuration and sets the input NTP servers in the configuration.
+      - If NTP based time synchronization is used internally, the NTP daemon will be restarted to reload given NTP configuration.
+      - In case NTP based time synchronization is not used, this method only replaces servers in the NTP configuration.
+    type: list
+    elements: str
+  noproxy:
+    description:
+      - List of hosts that should be ignored by proxy configuration.
+    type: list
+    elements: str
+  proxy:
+    elements: dict
+    type: list
+    description:
+      - A list of proxy configurations.
+    suboptions:
+        enabled:
+          description:
+            - Define if this proxy configuration should be enabled.
+          type: bool
+          required: true
+        url:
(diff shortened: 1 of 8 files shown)


mariolenz:
@machacekondra @bardielle I didn't have a closer look the the code, but could I suggest a different name for the module? `system` is pretty generic. It looks like this module configures the vCenter / vCSA, not any _guest_ or _(ESXi) host_ or any other system.

If I'm correct, I suggest to call the module `vcenter_system` or `vcsa_system` or similar. Additionally, I would update the description. At the moment it says `Configure system settings of VMware appliance` but there are a lot of VMware appliances. If the module is about the vCenter appliance, as I think it is, it should state this clearly.

Sorry if I understand the module wrong. I didn't have the time for a closer look.

bardielle:
@mariolenz I agree with you it will be more clear.

machacekondra:
@mariolenz What about call it `appliance_settings` or something? I don't know much of the VMware terminology, so would be great to hear your advice. The module basically configure anything which is exposed by vcenter on port 5480. So I am open to any of your suggestions.

mariolenz:
@machacekondra I would say that virtual [appliance](https://en.wikipedia.org/wiki/Virtual_appliance) (often delivered in form of an [OVF](https://en.wikipedia.org/wiki/Open_Virtualization_Format)) is still too generic. There are both different virtual appliances from VMware (like Aria Operations, if it's still called that) and others like Cisco, HPE, Dell, NetApp... ESXi itself has an appliance-like quality, too. Although it's not a _virtual_ appliance, usually.

So maybe call it `vcsa_settings`? vCSA is the official name and stands for _vCenter Server Appliance_ with vCenter Server being the original software to manage ESXi hosts and had to be installed on a Windows Server. The vCSA is, well, the SW delivered in form of a virtual appliance. That is, including an OS. Photon OS, I think :thinking: 

BTW I think port 5480 often is called the VAMI (Virtual Appliance Management Interface, or something very similar).

@ihumster What do you think? It's easy to fix a description in the documentation but hard to rename a module. So we should try and have a good name from the beginning.

ihumster:
Yep, I think `vcsa` (how abbreviation denoting vCenter Server Appliance) a good name to designate a module that configures appliance.

Modules starts with `vcenter_` will configure something inside vCenter Server.

plugins/modules/vcsa_settings.py
```diff
@@ -0,0 +1,576 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: vcsa_settings
+short_description: Configure vCenter Server Appliance settings
+version_added: 1.2.0
```
bardielle, line 16: This will be 2.0.0

bardielle:
@mariolenz can we merge this PR? Or do you have any other comments?

plugins/modules/vcsa_settings.py
```diff
@@ -0,0 +1,576 @@
...
+    choices:
+      - is_static
+      - dhcp
+    description:
+      - Set the DNS mode.
+    type: str
+  dns_append:
+    description:
+      - If I(true) items from C(dns_domains) and C(dns_servers) will be added to already configured DNS domains/servers.
+      - If I(false) domains/servers will be overriden.
```
mariolenz, line 45:
```suggestion
      - If V(true) items from O(dns_domains) and O(dns_servers) will be added to already configured DNS domains/servers.
      - If V(false) domains/servers will be overriden.
```

plugins/modules/vcsa_settings.py
```diff
@@ -0,0 +1,576 @@
...
+      - Enable/Disable state of Direct Console User Interface (DCUI TTY2).
+    type: bool
+  shell_enabled:
+    description:
+      - Enable/Disable state of BASH, that is, access to BASH from within the controlled CLI.
+    type: bool
+  shell_timeout:
+    description:
+      - The timeout (in seconds) specifies how long you enable the Shell access. The maximum timeout is 86400 seconds(1 day).
+      - This parameter is mandatory in case C(shell_enabled) is I(true).
```
mariolenz, line 127:
```suggestion
      - This parameter is mandatory in case O(shell_enabled=true).
```

plugins/modules/vcsa_settings.py
```diff
@@ -0,0 +1,576 @@
...
+          - Defines firewall rule policies.
+        type: str
+        choices:
+          - ACCEPT
+          - IGNORE
+          - REJECT
+          - RETURN
+      interface_name:
+        description:
+          - The interface to which this rule applies. An I(*) indicates that the rule applies to all interfaces.
```
mariolenz, line 163:
```suggestion
        choices:
          - ACCEPT
          - IGNORE
          - REJECT
          - RETURN
          - *
      interface_name:
        description:
          - The interface to which this rule applies. An V(*) indicates that the rule applies to all interfaces.
```

plugins/modules/vcsa_settings.py
```diff
@@ -0,0 +1,576 @@
...
+            shell_enabled=dict(type='bool'),
+            shell_timeout=dict(type='int'),
+            consolecli_enabled=dict(type='bool'),
+            dcui_enabled=dict(type='bool'),
+            firewall_rules_append=dict(type='bool', default=True),
+            firewall_rules=dict(type='list', elements='dict', options=dict(
+                address=dict(type='str'),
+                prefix=dict(type='int'),
+                interface_name=dict(type='str'),
+                policy=dict(type='str', choices=['ACCEPT', 'IGNORE', 'REJECT', 'RETURN']),
```
mariolenz, line 539:
```suggestion
                policy=dict(type='str', choices=['ACCEPT', 'IGNORE', 'REJECT', 'RETURN', '*']),
```

mariolenz (commented):
Two suggestions:

1. Use [Semantic markup](https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation)
2. If `*` is a valid option for `interface_name`, I think it has to be defined in argument_spec.

machacekondra:
> I suggest to use [Semantic markup](https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation) in the documentation.

Thank you, Mario.

mariolenz (approved):
LGTM

PR #24, opened 2024-05-21
prepare vcsim role - using podman instead of docker while pulling vcs…

bardielle:
…im image

##### SUMMARY
 Use podman for pulling the vcsim image in the integration role prepare_vcsim

##### ISSUE TYPE
- Test fixes

##### ADDITIONAL INFORMATION
Right now there is a new[ issue](https://github.com/psf/requests/issues/6707) using docker.

### DIFF
--- a/tests/integration/targets/prepare_vcsim/tasks/main.yml
+++ b/tests/integration/targets/prepare_vcsim/tasks/main.yml
@@ -2,20 +2,21 @@
 - name: Install dependencies
   ansible.builtin.pip:
     name:
-      - pyvmomi
       - requests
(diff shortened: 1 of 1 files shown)


changelogs/fragments/24-use-podman-in-integration-test.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
```
machacekondra, line 1: Please remove this changelog, as it's related to tests not to code. Let's keep in changelog only things related to whatever is released to user.

machacekondra (approved):

PR #25, opened 2024-05-21
Deprecating the old module vm_list_group_by_clusters

bardielle:
##### SUMMARY
Deprecating the module vm_list_group_by_clusters since it will be replaced with vm_list_group_by_clusters_info in v2.0.0

##### ISSUE TYPE
Adding a deprecating 

##### COMPONENT NAME
Deprecating the module vm_list_group_by_clusters since it will be replaced with vm_list_group_by_clusters_info in v2.0.0 and since then we will have 2 modules. 
I used that [doc](https://docs.ansible.com/ansible/latest/dev_guide/module_lifecycle.html#changing-a-module-or-plugin-name-in-the-ansible-main-repository) in order to sign that module as deprecated

### DIFF
--- a/changelogs/changelog.yaml
+++ b/changelogs/changelog.yaml
@@ -14,7 +14,7 @@ releases:
     release_date: '2024-05-07'
   1.1.0:
     changes:
-      major_changes:
+      minor_changes:
       - Added module vm_list_group_by_clusters
     fragments:
     - 16-vm_list_group_by_clusters.yml
--- a/changelogs/fragments/25-deprecated_module_vm_list_goup_by_clusters.yml
+++ b/changelogs/fragments/25-deprecated_module_vm_list_goup_by_clusters.yml
@@ -1,0 +1,2 @@
+deprecated_features:
+  - vm_list_group_by_clusters - deprecate the module since it was renamed to `vm_list_group_by_clusters_info`
--- a/changelogs/fragments/25-deprecated_module_vm_list_goup_by_clusters.yml
+++ b/changelogs/fragments/25-deprecated_module_vm_list_goup_by_clusters.yml
@@ -1,2 +1,2 @@
 deprecated_features:
-  - vm_list_group_by_clusters - deprecate the module since it was renamed to `vm_list_group_by_clusters_info`
+  - vm_list_group_by_clusters - deprecate the module since it was renamed to ``vm_list_group_by_clusters_info``
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,2 +1,9 @@
 ---
 requires_ansible: '>=2.15.0'
+plugin_routing:
+    modules:
+        vm_list_group_by_clusters:
+            redirect: vmware.vmware.vm_list_group_by_clusters_info
+            deprecation:
+                removal_version: 2.0.0
+                warning_text: Use M(vmware.vmware.vm_list_group_by_clusters_info) instead.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ vmware.vmware Release Notes
 v1.1.0
 ======
 
-Major Changes
+Minor Changes
 -------------
 
 - Added module vm_list_group_by_clusters


changelogs/changelog.yaml
```diff
@@ -14,7 +14,7 @@ releases:
     release_date: '2024-05-07'
   1.1.0:
     changes:
-      major_changes:
+      minor_changes:
       - Added module vm_list_group_by_clusters
```
mariolenz, line 18: I'm not sure if it's acceptable to alter the changelog after the release. Not 100% sure, though.

cc @Andersson007

meta/runtime.yml
```diff
@@ -1,2 +1,9 @@
 ---
 requires_ansible: '>=2.15.0'
+plugin_routing:
+    modules:
+        vm_list_group_by_clusters:
+            redirect: vmware.vmware.vm_list_group_by_clusters_info
+            deprecation:
+                removal_version: 2.0.0
+                warning_text: Use vmware.vmware.vm_list_group_by_clusters_info instead.
```
mariolenz, line 9: I haven't renamed a module for quite some time. But since you're `plugin_routing` here, do we _also_ need `plugins/modules/_vm_list_group_by_clusters.py`? I should say no, but I honestly don't know :shrug:

mariolenz:
Just wanted to work on this, also :laughing:

meta/runtime.yml
bardielle: As far as I understood from that [page](https://docs.ansible.com/ansible/latest/dev_guide/module_lifecycle.html#changing-a-module-or-plugin-name-in-the-ansible-main-repository) we need but I'm not sure
 @Andersson007 what do you think?

bardielle:
@mariolenz thank you for your comments :)

meta/runtime.yml
Andersson007: I don't know two, i would look in other collections how it's done there. Maybe would switch between branches in community.general and see

changelogs/changelog.yaml
bardielle: Thank I updated that file as well

meta/runtime.yml
bardielle: So it seems that it's not recommended anymore to add the "_" prefix, Also I got some errors regarding that when I used to add the "_vm_list_group_by_clusters" instead of "vm_list_group_by_clusters"

felixfontein: Underscores in collections do *not* deprecate. That was never really the case.

changelogs/changelog.yaml
felixfontein: CHANGELOG.rst is likely autogenerated, you can use `antsibull-changelog generate` to update it from `changelogs/changelog.yaml`.

It's best to not manually edit the output file since antsibull-changelog will overwrite it for the next release anyway. Better only change the YAML file (either directly or indirectly through `antsibull-changelog release`) and let antsibull-changelog update the output files.

(You can also enable MarkDown output for antsibull-changelog BTW.)

plugins/modules/vm_list_group_by_clusters.py
felixfontein, line 1: Why are you adding this file to the repository? It will never be used anyway due to the redirect.

bardielle: this one is an old version of `vm_list_group_by_clusters_info.py` that I removed last PR but it was already [released](https://galaxy.ansible.com/ui/repo/published/vmware/vmware/content/module/vm_list_group_by_clusters/)

We prefer to release a minor version with deprecated module then release a major version

meta/runtime.yml
bardielle: I'm agree with you but I followed after the [documentation](https://docs.ansible.com/ansible/latest/dev_guide/module_lifecycle.html#the-lifecycle-of-an-ansible-module-or-plugin ) : 

> Rename the file so it starts with an _, for example, rename old_cloud.py to _old_cloud.py. This keeps the module available and marks it as deprecated on the module index pages. 

Also, this is a[ PR](https://github.com/ansible/ansible/pull/43781/files) example that was added to the documentation and it seems to include the underscore.

bardielle: I'm resolving those comments since we agree not to add underscore.

plugins/modules/vm_list_group_by_clusters.py
felixfontein: But this code will never actually be executed, and not even be used by ansible-doc, since there's a redirect in meta/runtime.yml for it.

With that redirect, this file should be deleted. The last Ansible version which didn't knew about redirects was Ansible 2.9, and that's explicitly not supported by this collection.

meta/runtime.yml
felixfontein: The beginning of the section you are quoting (https://docs.ansible.com/ansible/latest/dev_guide/module_lifecycle.html#deprecating-modules-and-plugins-in-the-ansible-main-repository) says "To deprecate a module **in ansible-core,** you must:" (emphasis mine). The section after that is for deprecating a module or plugin in a collection, and it doesn't mention underscores.

plugins/modules/vm_list_group_by_clusters.py
bardielle: I removed it thanks...

changelogs/changelog.yaml
bardielle: Thanks @felixfontein I reverted the manual change and ran `antsibull-changelog generate` and pushed it.

bardielle:
@mariolenz @felixfontein @Andersson007 Any disclaimers regarding that PR?

mariolenz (approved):
LGTM

changelogs/fragments/25-deprecated_module_vm_list_goup_by_clusters.yml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - vm_list_group_by_clusters - deprecate the module since it was renamed to `vm_list_group_by_clusters_info`
```
felixfontein, line 2:
```suggestion
  - vm_list_group_by_clusters - deprecate the module since it was renamed to ``vm_list_group_by_clusters_info``
```

felixfontein (commented):
One last detail- changelog fragments are using ReStructured Text, so you need double backticks.

bardielle:
thank you @felixfontein

felixfontein (approved):

PR #26, opened 2024-05-23
Add content_template module

machacekondra:
Module to create template in content library from virtual machine.

### DIFF
--- a/plugins/module_utils/vmware_rest_client.py
+++ b/plugins/module_utils/vmware_rest_client.py
@@ -21,25 +21,15 @@ except ImportError:
     REQUESTS_IMP_ERR = traceback.format_exc()
     HAS_REQUESTS = False
 
-PYVMOMI_IMP_ERR = None
-try:
-    from pyVim import connect  # noqa: F401, pylint: disable=unused-import
-    from pyVmomi import vim  # noqa: F401, pylint: disable=unused-import
-    HAS_PYVMOMI = True
-except ImportError:
-    PYVMOMI_IMP_ERR = traceback.format_exc()
-    HAS_PYVMOMI = False
-
 VSPHERE_IMP_ERR = None
 try:
-    from com.vmware.vapi.std_client import DynamicID
     from vmware.vapi.vsphere.client import create_vsphere_client
     from com.vmware.vapi.std.errors_client import Unauthorized
     from com.vmware.content.library_client import Item
     from com.vmware.vcenter_client import (Folder,
                                            Datacenter,
                                            ResourcePool,
-                                           Datastore,
+                                           VM,
                                            Cluster,
                                            Host,
                                            VM)
@@ -97,9 +87,6 @@ class VmwareRestClient(object):
         if not HAS_REQUESTS:
             self.module.fail_json(msg=missing_required_lib('requests'),
                                   exception=REQUESTS_IMP_ERR)
-        if not HAS_PYVMOMI:
-            self.module.fail_json(msg=missing_required_lib('PyVmomi'),
-                                  exception=PYVMOMI_IMP_ERR)
         if not HAS_VSPHERE:
             self.module.fail_json(
                 msg=missing_required_lib('vSphere Automation SDK',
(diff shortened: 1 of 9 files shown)


plugins/module_utils/vmware_rest_client.py
```diff
@@ -21,25 +21,15 @@
     REQUESTS_IMP_ERR = traceback.format_exc()
     HAS_REQUESTS = False
 
-PYVMOMI_IMP_ERR = None
-try:
```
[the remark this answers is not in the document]
machacekondra: It may, but we agreed to start with only relevant needed functionality, same for the other module_utils files. We would need to fix it there as well, but on forum we agreed on this approach.

plugins/module_utils/vmware_rest_client.py
```diff
@@ -381,178 +220,121 @@ def get_datacenter_by_name(self, datacenter_name):
...
-        return folder
+        return folder_summaries[0].folder if len(folder_summaries) > 0 else None
 
-    def get_resource_pool_by_name(self, datacenter_name, resourcepool_name, cluster_name=None, host_name=None):
+    def get_resource_pool_by_name(self, resourcepool_name, datacenter_name=None, cluster_name=None, host_name=None):
         """
         Returns the identifier of a resource pool
         with the mentioned names.
         """
         datacenter = self.get_datacenter_by_name(datacenter_name)
```
bardielle, line 250: those 2 lines :
 datacenter = self.get_datacenter_by_name(datacenter_name)
 datacenters = set([datacenter]) if datacenter else set()

 are repeated at  least 5 times - can you replace the function to be get_datasenter_set_by_names?

plugins/modules/content_create_template.py
```diff
@@ -0,0 +1,203 @@
...
+        folder=dict(type='str'),
+    )
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        required_one_of=[('host', 'resource_pool', 'cluster')],
+    )
+
+    result = {'failed': False, 'changed': False}
+    vmware_contentlib_create = VmwareContentCreateTemplate(module)
```
bardielle, line 190: what about deleting and updating template?

plugins/modules/content_create_template.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
```
bardielle, line 2: I think module name should be content_template and handle create, delete and update as well

plugins/modules/content_create_template.py
machacekondra: Added deleting, updating support only notes and name, which is not really needed imho.

plugins/modules/content_template.py
```diff
@@ -0,0 +1,236 @@
...
+        self.cluster = self.params.get('cluster')
+        self.resource_pool = self.params.get('resource_pool')
+        self.library = self.params.get('library')
+        self.folder = self.params.get('folder')
+        self.vm_name = self.params.get('vm_name')
+
+    def create_template_from_vm(self):
+        template = self.get_library_item_from_content_library_name(self.template, self.library)
+        if template:
+            self.result['template_info'] = dict(
```
bardielle, line 144: This will be an error? I think we should throw an error if it's update case

machacekondra: This shouldn't be error. This should return `changed=False` and info that the template already exists.

bardielle: if this is the same object I agree with you, but what about editing some fields?

machacekondra: What fields do you mean? Name can be updated, and in future, would be nice to support it.

bardielle: I mean if user will apply changes in name or library it will not change nothing but will response that everything is good

mikemorency: If the user changes something in the placement spec but keeps the name that same, I think this module will give them a confusing result. It will return OK when in reality the template does not have the configuration they expect

plugins/module_utils/vmware_rest_client.py
mikemorency, line 1: just to clarify, the large deletions in this file are just general cleanup right? Not related to this module directly

plugins/modules/content_template.py
machacekondra: As far as I understand the placement is used only for the import process. So after import is done, it really doesn't make sense to change the field. So IMHO it's better to succed, as nothing was really changed then to fail, but makes sense to document it well, so users understand.

plugins/module_utils/vmware_rest_client.py
machacekondra: Oh, yes, I found out, there are methods, which shouldn't be there, I guess I forgot to remove them before.

mikemorency (approved):

bardielle (approved):

bardielle:
@machacekondra fix the rebase-conflicts and merge your PR  :)

PR #36, opened 2024-06-03
[WIP] vm_list_group_by_clusters: Fix return value

mariolenz:
##### SUMMARY
This is a follow-up to #25

The problem is that the new module changes the return value. It mustn't, at least not when called with the old name.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vm_list_group_by_clusters
vm_list_group_by_clusters_info

##### ADDITIONAL INFORMATION
[Include vmware.vmware collection into Ansible package comment](https://github.com/ansible-collections/ansible-inclusion/discussions/75#discussioncomment-9640994)


mariolenz:
I think this has been fixed meanwhile in #37

PR #37, opened 2024-06-05
Add returned object vm_list_group_by_clusters if deprecated module was…

bardielle:
… called

##### SUMMARY
Fixing Felix's comment from this [discussion](https://github.com/ansible-collections/ansible-inclusion/discussions/75#discussioncomment-9640996).
Add a return value's name of the deprecated module in case the module is called with the old name.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vm_list_group_by_clusters_info

### DIFF
--- a/plugins/modules/vm_list_group_by_clusters_info.py
+++ b/plugins/modules/vm_list_group_by_clusters_info.py
@@ -191,7 +191,12 @@ def main():
 
     vmware_vm_list_group_by_clusters_mgr = VmwareVMList(module)
     vm_list_group_by_clusters_info = vmware_vm_list_group_by_clusters_mgr.get_vm_list_group_by_clusters()
-    module.exit_json(changed=False, vm_list_group_by_clusters_info=vm_list_group_by_clusters_info)
+    # Till we will release the next major version 2.0.0 we should keep the deprecated module return value
+    if not module._name.endswith('_info'):
+        module.exit_json(changed=False, vm_list_group_by_clusters_info=vm_list_group_by_clusters_info,
+                         vm_list_group_by_clusters=vm_list_group_by_clusters_info)
+    else:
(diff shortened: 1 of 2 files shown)


plugins/modules/vm_list_group_by_clusters_info.py
```diff
@@ -191,7 +191,12 @@ def main():
 
     vmware_vm_list_group_by_clusters_mgr = VmwareVMList(module)
     vm_list_group_by_clusters_info = vmware_vm_list_group_by_clusters_mgr.get_vm_list_group_by_clusters()
-    module.exit_json(changed=False, vm_list_group_by_clusters_info=vm_list_group_by_clusters_info)
+    # Till we will release the next major version 2.0.0 we should keep the deprecated module return value
+    if not module._name.endswith('_info'):
+        module.exit_json(changed=False, vm_list_group_by_clusters_info=vm_list_group_by_clusters_info)
```
machacekondra, line 196: If I am not mistaken, this should be:
```
module.exit_json(changed=False, vm_list_group_by_clusters=vm_list_group_by_clusters_info
```

bardielle: Fixed it
thanks

machacekondra: Looks still the same for me

machacekondra: Ah I see, you remove the `not`.

machacekondra: Actually, it's imho wrong. Because what you want is to keep backward compatibility. That means, you want to return same value for `_info` + old value. But for old module, you just keep returning what you did.

bardielle: OK I changed it

mikemorency (approved):

machacekondra (approved):

PR #39, opened 2024-06-18
Mm feature/save vm as template

mikemorency:
##### SUMMARY
Adding module to create a template from an existing VM and saving the template in a folder in VCenter. If the template already exists, the module exits without changing anything.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
folder_template

### DIFF
--- a/changelogs/fragments/39-add_folder_template.yml
+++ b/changelogs/fragments/39-add_folder_template.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - folder_template - add module and tests to create a template from an existing VM in vcenter and store the template in a folder
--- a/plugins/module_utils/vmware.py
+++ b/plugins/module_utils/vmware.py
@@ -58,6 +58,12 @@ def vmware_argument_spec():
                       required=False,
                       no_log=True,
                       fallback=(env_fallback, ['VMWARE_PASSWORD'])),
+        cluster=dict(type='str',
+                     aliases=['cluster_name'],
+                     required=False),
+        datacenter=dict(type='str',
+                        aliases=['datacenter_name'],
+                        required=False),
         port=dict(type='int',
                   default=443,
                   fallback=(env_fallback, ['VMWARE_PORT'])),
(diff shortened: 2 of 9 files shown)


changelogs/fragments/39-add_folder_template.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
```
bardielle, line 1: Does it support storing it only in a folder?

plugins/module_utils/vmware.py
```diff
@@ -218,3 +224,41 @@ def is_vcenter(self):
...
+        """
+        obj = None
+        container = self.content.viewManager.CreateContainerView(self.content.rootFolder, vimtype, True)
+        for c in container.view:
+            if c.name == name:
+                obj = c
+                break
+        return obj
+
+    def get_vm_by_name(self, vm_name, fail_on_missing=False):
```
bardielle, line 242: vm name is not unique. I think you should provide the root folder you're looking into
also, what is the strategy if there are multiple VMs with the same name?
For example, in community.vmware.vmware_guest_info there is the property `name_match`:
```
   name_match:
     description:
     - If multiple VMs matching the name, use the first or last found
     default: 'first'
     choices: ['first', 'last']
     type: str
```

plugins/module_utils/vmware.py
```diff
@@ -218,3 +224,41 @@ def is_vcenter(self):
...
+        vm = self.__get_obj_by_name([vim.VirtualMachine], vm_name)
+        if not vm and fail_on_missing:
+            self.module.fail_json("Unable to find VM with name %s" % vm_name)
+        return vm
+
+    def get_folder_by_name(self, folder_name, fail_on_missing=False):
+        folder = self.__get_obj_by_name([vim.Folder], folder_name)
```
bardielle, line 249: You should get the parent folder 
If I'm not wrong, folder name is not unique

plugins/modules/folder_template.py
```diff
@@ -0,0 +1,210 @@
...
+    description:
+      - The name of the vm to be used to create the template
+    type: str
+    required: True
+  folder:
+    description:
+      - The name of the folder that the new template should be placed in
+    type: str
+    required: True
+  name:
```
bardielle, line 34: I think template name will be more clear 
and for modification - 
new_template_name (not sure if it's supported)

machacekondra (commented):
What is difference between this module and: https://docs.ansible.com/ansible/latest/collections/community/vmware/vmware_guest_module.html#parameter-is_template

changelogs/fragments/39-add_folder_template.yml
mikemorency: Yes, I think the other option is storing it in a content library. Even if the user didnt specify a folder path, it would technically be stored in the `/<datacenter>/vms` folder

mikemorency:
@machacekondra I didnt know that feature existed. I think the only difference is that converts a VM into a template, where-as this clones the VM and creates a template from the clone. So you keep the VM

It might be useful to keep this around if we want to deconstruct the vmware_guest module in the future?

changelogs/fragments/39-add_folder_template.yml
mariolenz: "Normal" templates are quite old, content libraries are much younger. AFAIK there are significant differences between them, both technically and how people use them.

PR #42, opened 2024-06-25
ACA-1650: Rename get_vm_by_name method

machacekondra:
This method is used twice in VmwareRestClient class.


mikemorency (approved):

mikemorency:
the two methods seem very similar, maybe we should look into consolidating them

machacekondra:
> the two methods seem very similar, maybe we should look into consolidating them

Yeah, I think we should keep one or the other and change the code, which calls them.

bardielle:
Why can't we use the original REST method get_vm_by_name? 
What si the different between those two methods?
Can you brink some more details and add a comment on top of that method?

PR #48, opened 2024-07-10
Mm feature/save vm as template

mikemorency:
##### SUMMARY
This module creates a template in a folder from a VM. Its designed to replace some of the functionality in community.vmware.vmware_guest.

The module allows a user to create or delete a template in a folder. If you create a template, you must specify a VM to clone and use as a base for the template. Once the template is created, the user would use another module to modify the configuration (right now I think they would use vmware_guest, but a new module would probably be useful).

This module is similar in functionality to the existing content_template module, which creates/deletes a template in a content library.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
folder_template_from_vm

### DIFF
--- a/changelogs/fragments/48-add_folder_template.yml
+++ b/changelogs/fragments/48-add_folder_template.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - folder_template_from_vm - add module and tests to create a template from an existing VM in vcenter and store the template in a folder
--- a/plugins/doc_fragments/vmware.py
+++ b/plugins/doc_fragments/vmware.py
@@ -46,6 +46,16 @@ options:
       - If the value is not specified in the task, the value of environment variable E(VMWARE_PORT) will be used instead.
       type: int
       default: 443
+    datacenter:
+      description:
+      - The datacenter to use when connecting to a vCenter.
+      type: str
+      aliases: [ datacenter_name ]
+    cluster:
+      description:
+      - The cluster to use when connecting to a vCenter.
+      type: str
+      aliases: [ cluster_name ]
     proxy_host:
       description:
       - Address of a proxy that will receive all HTTPS requests and relay them.
@@ -95,6 +105,16 @@ options:
       - If the value is not specified in the task, the value of environment variable E(VMWARE_PORT) will be used instead.
       type: int
       default: 443
+    datacenter:
+      description:
+      - The datacenter to use when connecting to a vCenter.
+      type: str
+      aliases: [ datacenter_name ]
+    cluster:
+      description:
+      - The cluster to use when connecting to a vCenter.
+      type: str
+      aliases: [ cluster_name ]
     proxy_host:
       description:
       - Address of a proxy that will receive all HTTPS requests and relay them.
--- a/plugins/module_utils/vmware.py
+++ b/plugins/module_utils/vmware.py
@@ -58,6 +58,12 @@ def vmware_argument_spec():
                       required=False,
                       no_log=True,
                       fallback=(env_fallback, ['VMWARE_PASSWORD'])),
+        cluster=dict(type='str',
+                     aliases=['cluster_name'],
+                     required=False),
+        datacenter=dict(type='str',
+                        aliases=['datacenter_name'],
+                        required=False),
         port=dict(type='int',
                   default=443,
                   fallback=(env_fallback, ['VMWARE_PORT'])),
@@ -225,7 +231,9 @@ class PyVmomi(object):
 
     def get_objs_by_name_or_moid(self, vimtype, name, return_all=False):
         """
-        Get any vsphere object associated with a given text name and vim type.
+        Get any vsphere objects associated with a given text name or MOID and vim type.
+        Different objects have different unique-ness requirements for the name parameter, so
+        you may get one or more objects back. The MOID should always be unique
         Args:
             vimtype: The type of object to search for
             name: The name or the ID of the object to search for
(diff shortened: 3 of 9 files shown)


tests/integration/targets/vmware_folder_template_from_vm/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
+vcenter_port: 443
+run_on_simulator: false
+
+template_folder: e2e-qe
+
+vm: test-folder_template
+vm_cluster: "Eco-Cluster"
+vm_datacenter: "Eco-Datacenter"
```
shellymiron, line 8: just fyi we have a [common vars yml file ](https://github.com/ansible-collections/vmware.vmware/blob/main/tests/integration/targets/group_vars.yml)you can import and use instead of writing the names

tests/integration/targets/vmware_folder_template_from_vm/vars.yml
```diff
@@ -0,0 +1,5 @@
+---
```
elsapassaro, line 1: The file with variables for real vcenter should be in `defaults/main.yml` and the variables for the simulator in vars.yml

mikemorency: ah ok. is there an existing example of this? i tried looking at the other roles but they all seemed a bit different so i think im getting confused. Seems like `vmware_appliance_info` does what your talking about?

elsapassaro: yes, `vmware_appliance_info` does this, and actually I think all other integration targets running in upstream/downstream CI (all the ones prefixed with `vmware_`)

changelogs/fragments/48-add_folder_template.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - folder_template_from_vm - add module and tests to create a template from an existing VM in vcenter and store the template in a folder
```
machacekondra, line 2: Those kind of things usually read users not developers. I think users are not interested about if you added tests. Developers are, but developers read git commit message, so I think it's not necessary to note about tests in changelog.

plugins/module_utils/vmware.py
```diff
@@ -218,3 +224,141 @@ def is_vcenter(self):
...
+    def get_vm_by_name(self, vm_name, fail_on_missing=False, name_match=None):
+        """
+            Get the vms matching the given name. VM names may not be unique
+            in a given cluster.
+            Args:
+                vm_name: Name of the VM to search for
+                fail_on_missing: If true, an error will be thrown if no VMs are found
+                name_match: If provided, return one VM from the list of VMs. Either 'first' or 'last'
+            Returns:
+                list(vm), vm if name_match is provided, or None if no matches were found
```
machacekondra, line 258: It may also return just string in case `name_match` is provided. Might be bit confusing. I think better to return always a list or a string.

plugins/modules/folder_template_from_vm.py
```diff
@@ -0,0 +1,321 @@
...
+
+        return template_location_spec
+
+    def __wait_for_template(self, task):
+        """
+        Waits and provides updates on a vSphere task
+        """
+        while task.info.state == vim.TaskInfo.State.running:
+            time.sleep(2)
+
```
machacekondra, line 260: I think easier to read is:
```
if task.info.state != vim.TaskInfo.State.success:
  self.module.fail_json(msg=task.info.error)
```

plugins/modules/folder_template_from_vm.py
```diff
@@ -0,0 +1,321 @@
...
+        if self.resource_pool:
+            template_location_spec.pool = self.get_resource_pool_by_name(self.resource_pool, fail_on_missing=True)
+
+        return template_location_spec
+
+    def __wait_for_template(self, task):
+        """
+        Waits and provides updates on a vSphere task
+        """
+        while task.info.state == vim.TaskInfo.State.running:
```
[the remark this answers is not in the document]
mikemorency: yea actually as part of another module im migrating, i noticed it was also using a similar function. So ive built out a class/function to do this.
When i open that PR, ill migrate this module to use that shared code

plugins/module_utils/vmware.py
mikemorency: i think it should be a list of VM objects instead of just a string since the object is generally more useful. But always returning a list does sound less confusing, ill implement that!

changelogs/fragments/48-add_folder_template.yml
mikemorency: ah ok, makes sense! ill keep that in mind for the future

plugins/doc_fragments/vmware.py
```diff
@@ -46,6 +46,16 @@ class ModuleDocFragment(object):
       - If the value is not specified in the task, the value of environment variable E(VMWARE_PORT) will be used instead.
       type: int
       default: 443
+    datacenter:
+      description:
```
bardielle, line 50: Does it a required parameter?

mikemorency: for this module yes it is because we need to get the absolute folder path in order to save the template. And the folder path is <datacenter>/vms/whatever

plugins/doc_fragments/vmware.py
```diff
@@ -46,6 +46,16 @@ class ModuleDocFragment(object):
...
+      - The datacenter to use when connecting to a vCenter.
+      type: str
+      aliases: [ datacenter_name ]
+    cluster:
+      description:
```
bardielle, line 55: Is it a required parameter?

plugins/doc_fragments/vmware.py
bardielle: I found it later... Resolving this comment

plugins/module_utils/vmware.py
```diff
@@ -218,3 +224,141 @@ def is_vcenter(self):
...
+        """
+        objs = []
+        container = self.content.viewManager.CreateContainerView(self.content.rootFolder, vimtype, True)
+        for c in container.view:
+            if c.name == name:
+                if first_only:
+                    return c
+                objs += [c]
+
+        return objs if objs else None
```
bardielle, line 247:
```suggestion
    def __get_objs_by_name(self, vimtype, name_or_id, first_only=False):
        """
        Get any vsphere object associated with a given text name or ID and vim type.
        Args:
            vimtype: The type of object to search for
            name_or_id: The name or the ID of the object to search for
            first_only: If true, return once the first object is found.
                        Useful when names must be unique
        Returns:
            list(object) or list() if no matches are found
        """
        obj = list()
        container = self.content.viewManager.CreateContainerView(
            self.content.rootFolder, vimtype, True)

        for c in container.view:
            if name_or_id in [c.name, c._GetMoId()]:
                if return_all is False:
                    return c
                else:
                    obj.append(c)

        if len(obj) > 0:
            return obj
        else:
            # for backwards-compat
            return None
```

plugins/module_utils/vmware.py
```diff
@@ -218,3 +224,141 @@ def is_vcenter(self):
...
+        if vms and name_match:
+            if name_match == 'first':
+                return [vms[0]]
+            elif name_match == 'last':
+                return [vms[-1]]
+            else:
+                self.module.fail_json("Unrecognized name_match option '%s' in get_vm_by_name method" % name_match)
+
+        return vms
+
```
bardielle, line 272:
```suggestion
    def get_vm(self, vm_name_or_id, fail_on_missing=False, name_match=None):
        """
            Get the vms matching the given name or ID. VM names may not be unique
            in a given cluster.
            Args:
                vm_name: Name or ID of the VM to search for
                fail_on_missing: If true, an error will be thrown if no VMs are found
                name_match: If provided, return list of one VM. Either 'first' or 'last'
            Returns:
                list(vm), or None if no matches were found
        """
        vms = self.__get_objs_by_name([vim.VirtualMachine], vm_name_or_id)
        if not vms and fail_on_missing:
            self.module.fail_json("Unable to find VM with name %s" % vm_name)
        if vms and name_match:
            if name_match == 'first':
                return [vms[0]]
            elif name_match == 'last':
                return [vms[-1]]
            else:
                self.module.fail_json("Unrecognized name_match option '%s' in get_vm" % name_match)

        return vms

```

plugins/module_utils/vmware.py
```diff
@@ -218,3 +224,141 @@ def is_vcenter(self):
...
+        container = self.content.viewManager.CreateContainerView(self.content.rootFolder, vimtype, True)
+        for c in container.view:
+            if c.name == name:
+                if first_only:
+                    return c
+                objs += [c]
+
+        return objs if objs else None
+
+    def get_vm_by_name(self, vm_name, fail_on_missing=False, name_match=None):
```
bardielle, line 249: why not using the rest client for get VM?

plugins/module_utils/vmware.py
mikemorency: that makes sense, ill make these changes!

plugins/module_utils/vmware.py
mikemorency: as far as i can tell, the rest client only allows you to search by name, not UUID or even MOID. if theres an example of that, it would be extremely helpful

tests/integration/targets/vmware_folder_template_from_vm/defaults/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+run_on_simulator: false
+template_folder: e2e-qe
+vm_name: folder_template_from_vm_test
```
elsapassaro, line 4: For each resource created by integration tests on our real vcenter we are using a `tiny_prefix` which is created for every run and is different for each integration target in order to prevent interference between tests. For upstream CI the tiny_prefix is defined [here](https://github.com/ansible-collections/vmware.vmware/blob/main/tests/integration/targets/runme.sh#L5), while for downstream CI it is a uuid created in the script that invokes the targets.

So I'd suggest the following change here:

```
vm_name: "{{ tiny_prefix }}_folder_template_from_vm"
```

plugins/module_utils/vmware.py
bardielle: But the rest is only by name so no need this func

bardielle (approved):

mariolenz (approved):
Quite a mouthful, but LGTM as far as I can see.

PR #50, opened 2024-07-17
migrate community.vmware.vmware_cluster_drs module

mikemorency:
##### SUMMARY
This is a migration of the community vmware_cluster_drs module to this collection. The inputs, actions, and results are all the same.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
cluster_drs

### DIFF
--- a/plugins/module_utils/vmware.py
+++ b/plugins/module_utils/vmware.py
@@ -228,7 +228,7 @@ class PyVmomi(object):
         elif api_type == 'HostAgent':
             return False
 
-    def get_objs_by_name_or_moid(self, vimtype, name, return_all=False):
+    def get_objs_by_name_or_moid(self, vimtype, name, return_all=False, search_root_folder=None):
         """
         Get any vsphere objects associated with a given text name or MOID and vim type.
         Different objects have different unique-ness requirements for the name parameter, so
@@ -238,12 +238,18 @@ class PyVmomi(object):
             name: The name or the ID of the object to search for
             return_all: If true, return all the objects that were found.
                         Useful when names must be unique
+            search_root_folder: The folder object that should be used as the starting point
+                                for searches. Useful for restricting search results to a
+                                certain datacenter (search_root_folder=datacenter.hostFolder)
         Returns:
             list(object) or list() if no matches are found
         """
+        if not search_root_folder:
+            search_root_folder = self.content.rootFolder
+
         obj = list()
         container = self.content.viewManager.CreateContainerView(
-            self.content.rootFolder, vimtype, True)
+            search_root_folder, vimtype, True)
 
         for c in container.view:
             if name in [c.name, c._GetMoId()]:
@@ -277,17 +283,6 @@ class PyVmomi(object):
             The distributed portgroup object
         """
         return self.get_objs_by_name_or_moid([vim.dvs.DistributedVirtualPortgroup], portgroup)
-
-    def get_datacenter_detailed(self, name):
-        """
-        Get detailed information about a datacenter
-        Args:
-            name: The name or the ID of the datacenter
-        Returns:
-            The datacenter object
-
-        """
-        return self.get_objs_by_name_or_moid([vim.Datacenter], name)
 
     def get_vm_using_params(
             self, name_param='name', uuid_param='uuid', moid_param='moid', fail_on_missing=False,
@@ -435,3 +430,44 @@ class PyVmomi(object):
         Get all virtual machines.
         """
         return self.list_all_objs_by_type([vim.VirtualMachine], folder=folder, recurse=recurse)
+
+    def get_datacenter_by_name(self, dc_name, fail_on_missing=False):
+        """
+            Get the datacenter matching the given name. Datacenter names must be unique
+            in a given vcenter, so only one object is returned at most.
+            Args:
+                dc_name: Name of the datacenter to search for
+                fail_on_missing: If true, an error will be thrown if no datacenters are found
+            Returns:
+                datacenter object or None
+        """
+        ds = self.get_objs_by_name_or_moid([vim.Datacenter], dc_name)
+        if not ds and fail_on_missing:
+            self.module.fail_json("Unable to find datacenter with name %s" % dc_name)
+        return ds
+
+    def get_cluster_by_name(self, cluster_name, fail_on_missing=False, datacenter=None):
+        """
+            Get the cluster matching the given name. Cluster names must be unique
+            in a given vcenter, so only one object is returned at most.
+            Args:
+                cluster_name: Name of the cluster to search for
+                fail_on_missing: If true, an error will be thrown if no clusters are found
+            Returns:
+                cluster object or None
+        """
+        search_folder = None
+        if datacenter and hasattr(datacenter, 'hostFolder'):
+            search_folder = datacenter.hostFolder
+
+        cluster = self.get_objs_by_name_or_moid(
+            [vim.ClusterComputeResource],
+            cluster_name,
+            return_all=False,
+            search_root_folder=search_folder
+        )
+
+        if not cluster and fail_on_missing:
+            self.module.fail_json("Unable to find cluster with name %s" % cluster_name)
+
+        return cluster
--- a/plugins/module_utils/vmware_type_utils.py
+++ b/plugins/module_utils/vmware_type_utils.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2023, Ansible Cloud Team (@ansible-collections)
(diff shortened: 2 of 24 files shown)


plugins/module_utils/vmware.py
```diff
@@ -281,3 +287,44 @@ def get_datacenter_detailed(self, name):
 
         """
         return self.get_objs_by_name_or_moid([vim.Datacenter], name)
+
+    def get_datacenter_by_name(self, dc_name, fail_on_missing=False):
```
mikemorency, line 291: @bardielle i added this method, but it similar to the one you added `get_datacenter_detailed`. I didnt see that method being used, so i was wondering if we should combine them?

bardielle, line 291: get_datacenter_by_name already exists in vmware_rest...
what is the difference between them?

plugins/module_utils/vmware_type_utils.py
```diff
@@ -0,0 +1,57 @@
...
+try:
+    from pyVmomi import vim, VmomiSupport
+    HAS_PYVMOMI = True
+except ImportError:
+    PYVMOMI_IMP_ERR = traceback.format_exc()
+    HAS_PYVMOMI = False
+
+
+def option_diff(options, current_options, truthy_strings_as_bool=True):
+    current_options_dict = {}
```
bardielle, line 19: I think the function name is not clear
also please add a comment

plugins/module_utils/vmware.py
mikemorency: it seems like the all of the `get_` rest methods (or most of them) return the object _identifier_ which is less useful to me. The methods in pyvmomi are returning the actual objects, which is what i need.

Maybe this is a topic better discussed as a group, @machacekondra , @mariolenz. Im not sure I understand the use case for getting an objects identifier instead of the object itself. That seems like its always less useful. 
Additionally, im not sure why we would use both the rest api vs pyvmomi or vice versa if they _can_ return the same infomation

mariolenz: > Im not sure I understand the use case for getting an objects identifier instead of the object itself. That seems like its always less useful.

I can't tell you. You see, a lot of the code has been written before my time. _Especially_ a lot of the code in the module utils. If we can simplify things here, let's do it!

Although we have to remember that the module utils (to the best of my knowledge) are considered part of the public API of the collection and breaking changes can only been done in a major release.

> Additionally, im not sure why we would use both the rest api vs pyvmomi or vice versa if they _can_ return the same infomation

A couple of years ago, VMware introduced the new REST API. There were some things possible via the new API, but not via the old (SOAP?) one. On the other hand, most or at least a lot of the stuff that was possible via the old API hasn't been implemented in the new one. Maybe that's the reason? I'll try to dig into this over the weekend, at least for this special case.

ATM I suggest to do as much as possible using pyVmomi because it's easier to install since it is available on PyPI. To the best of my knowledge, the REST SDK is harder to get.

bardielle: Yes I think we can combine them maybe we can filter out if returning detailed or only the moid

machacekondra: I agree, we can refactor the methods to return whatever will be more useful for us in future.

Not sure if module_utils is public API, it's not anywhere published and can't be used directly from playbooks, only internaly by modules, if I am not mistaken. That means if you change the code of the module, you can change the code of module_utils, without affecting users.

plugins/modules/cluster_drs.py
```diff
@@ -0,0 +1,258 @@
...
+    username: "{{ vcenter_username }}"
+    password: "{{ vcenter_password }}"
+    datacenter_name: DC0
+    cluster_name: "{{ cluster_name }}"
+    enable: true
+    drs_default_vm_behavior: partiallyAutomated
+  delegate_to: localhost
+'''
+
+RETURN = r'''
```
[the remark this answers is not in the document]
mariolenz: I don't think there's a reason to not document this. But [community.vmware.vmware_cluster_drs](https://github.com/ansible-collections/community.vmware/blob/6770ddf6879850ab37f828128c42bcaacb5fdac4/plugins/modules/vmware_cluster_drs.py) doesn't document the return, either.

So the question is: What do you want to do? Migrate the module as quick as possible and fix the return documentation later, or migrate the module and have the return documentation correct from the beginning?

plugins/module_utils/vmware.py
mariolenz: > Not sure if module_utils is public API, it's not anywhere published and can't be used directly from playbooks, only internaly by modules, if I am not mistaken. That means if you change the code of the module, you can change the code of module_utils, without affecting users.

To the best of my knowledge, playbooks can't use `module_utils`. But _other collections_ can. I'd like to re-use the module utils of this collection in [community.vmware](https://github.com/ansible-collections/community.vmware). So it _is_ part of the public API of your collection.

cc @felixfontein I'm not 100% sure here. I've tried to find some documentation about this, but I wasn't able to. Do you know if we've documented somewhere that module utils are part of a collections API and should follow SemVer? You're usually very well-informed, so maybe you could help here.

felixfontein: Generally module_utils are part of the public API of a collection (and thus should follow semver). The inclusion requirements have a section on module_utils that says how to mark them as private to a collection: https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_requirements.html#standards-for-developing-module-and-plugin-utilities

bardielle: cc @mikemorency

mikemorency: maybe we should mark them private for a release or two, until theres less changes? As we migrate more stuff from the community collection, I think we will want to address more issues like this one so minor releases would be nice
@mariolenz what do you think about that?

plugins/modules/cluster_drs.py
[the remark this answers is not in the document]
mikemorency: yea i copied what was already there. Ill update it though, no reason to wait until later

plugins/module_utils/vmware.py
```diff
@@ -279,17 +285,6 @@ def get_dvs_portgroup(self, portgroup):
...
-        """
-        Get detailed information about a datacenter
-        Args:
-            name: The name or the ID of the datacenter
-        Returns:
-            The datacenter object
-
-        """
-        return self.get_objs_by_name_or_moid([vim.Datacenter], name)
-
```
mariolenz, line 292: It looks like this function / method hasn't been released yet. So removing it wouldn't be a breaking change :partying_face: 

Could someone please double-check just to make sure?

plugins/module_utils/vmware.py
mariolenz: > maybe we should mark them private for a release or two, until theres less changes? As we migrate more stuff from the community collection, I think we will want to address more issues like this one so minor releases would be nice @mariolenz what do you think about that?

I think the original question from @bardielle was "get_datacenter_by_name already exists in vmware_rest... what is the difference between them?", wasn't it? Is this answered yet? The answer is different return values, right?

Since this is a _new_ function / method we don't run into a breaking change here. I just wanted to make you aware that you can't refactor the methods to return whatever will be more useful for us in future. At least not once it's been released.

I'm not sure if it's a good idea to mark this as private. It would make it easier for you to change it, but also impossible for other collections to re-use it. I'm really unsure... but since it's a new function / method, maybe you should mark it as private for now if you feel better with this.

> edit: Public -> Private requires a major release as well

Yes, but since it hasn't been released yet I don't see a problem there. Or am I wrong there and it has been released somewhere already?

plugins/module_utils/vmware.py
bardielle: Yep, it's fine removing it

bardielle (approved):
Overall looks good 
please get an agreement from @mariolenz before merging it

plugins/module_utils/vmware.py
mikemorency: So to summarize:

1. The original question about `get_datacenter_by_name` can be considered resolved. It is a duplicate method that has not been released yet and can be removed with a minor change.
2. The pyvmomi SDK should be prefered if both pyvmomi and the REST sdk provide the same functionality
3. Method names should be descriptive and consistent. Many methods in the community utils will need to be renamed/documented if they are moved to this collection
4. The module utils will be marked as private, temporarily, since theres likely going to be some changes in the next few minor releases related to takeaways 2 and 3. We would like to make it public after that, so community.vmware can leverage the new utils

@mariolenz please let me know if this sounds ok to you and if the PR looks good or needs any other improvements. Thanks to everyone for weighing in on this thread!

mariolenz (approved):
LGTM

PR #51, opened 2024-07-21
migrating community vm info modules into guest_info

mikemorency:
##### SUMMARY
This is an attempt to migrate vm info modules from community.vmware (vmware_guest_info, vmware_vm_info) into vmware.vmware. Since there is already a vm info module in vmware.vmware (guest_info), the functionality from the two community modules has been added to guest_info

This module also requires the rest and soap APIs. So for the simulator tests i added an API gateway so both APIs can be accessed on the same port.

All of the changes should be backwards compatible. We do plan on deprecating the module's old output and removing it at some point in the future. More on that in the additional info section.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
guest_info

##### ADDITIONAL INFORMATION
While the inputs for guest_info has not changed, the output should probably be changed. Theres some duplicated information, and some of the values previously returned by `guest_info` have similar names. 

Here is how the guest_info module currently returns info:
```
guests: [
  {
    old values here
  }
]
```

For now, we are returning the old values in two locations to the user:
```
guests: [
  {
    old and new values mixed together
    .....
    identity: {
        old values here
    }
  }
]
```

At some point in the future, the old values _not_ nested under `identity` will be removed. Resulting in:
It makes more sense to me if the old values are nested in an attribute like:
```
guests: [
  {
    new values here
    ......
    identity: {
        old values here
   }
  }
]
```

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -1,6 +1,6 @@
 ---
 name: Ansible Test
-on:  # yamllint disable-line rule:truthy
+on: # yamllint disable-line rule:truthy
   pull_request:
   merge_group:
     branches:
@@ -18,7 +18,10 @@ jobs:
         run: |
           set -euxo pipefail
           sudo apt-get update
-          sudo apt-get install podman
+
+          # recent version of podman has a bug. Fix is waiting deployment
+          # see https://github.com/actions/runner-images/issues/7753
+          sudo apt-get install podman=3.4.4+ds1-1ubuntu1 --allow-downgrades
 
       - name: Checkout repo
         uses: actions/checkout@v4
--- a/plugins/module_utils/vmware.py
+++ b/plugins/module_utils/vmware.py
@@ -405,7 +405,19 @@ class PyVmomi(object):
             self.module.fail_json("Unable to find resource pool with name %s" % pool_name)
         return pool
 
-    def get_all_objs_by_type(self, vimtype, folder=None, recurse=True):
+    def list_all_objs_by_type(self, vimtype, folder=None, recurse=True):
+        """
+            Returns a dictionary of all objects matching a given VMWare type.
+            You can also limit the search by folder and recurse if desired
+            Args:
+                vimtype: The type of object to search for
+                folder: vim.Folder, the folder object to use as a base for the search. If
+                        none is provided, the datacenter root will be used
+                recurse: If true, the search will recurse through the folder structure
+            Returns:
+                dicttionary of {obj: str}. The keys are the object while the values are the
+                object name
+        """
         if not folder:
             folder = self.content.rootFolder
 
@@ -422,4 +434,4 @@ class PyVmomi(object):
         """
         Get all virtual machines.
         """
-        return self.get_all_objs_by_type([vim.VirtualMachine], folder=folder, recurse=recurse)
+        return self.list_all_objs_by_type([vim.VirtualMachine], folder=folder, recurse=recurse)
--- a/plugins/module_utils/vmware.py
+++ b/plugins/module_utils/vmware.py
@@ -16,7 +16,6 @@ import atexit
 import ssl
 import traceback
 
-
 REQUESTS_IMP_ERR = None
 try:
     # requests is required for exception handling of the ConnectionError
@@ -294,7 +293,6 @@ class PyVmomi(object):
             self, name_param='name', uuid_param='uuid', moid_param='moid', fail_on_missing=False,
             name_match_param='name_match', use_instance_uuid_param='use_instance_uuid'):
         """
-            TODO
             Get the vms matching the common module params related to vm identification: name, uuid, or moid. Since
             MOID and UUID are unique identifiers, they are tried first. If they are not set, a search by name is tried
             which may give one or more vms.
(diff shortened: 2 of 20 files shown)


mariolenz:
@mikemorency Sorry, didn't find the time to review yet. Will try to do it tomorrow!

mariolenz (commented):
I wanted to check if this could really be a replacement for `comunity.vmware.vmware_guest_info` and at first glance it looks OK. I'd like to have a closer look, though, and also would want to compare it with `comunity.vmware.vmware_vm_info`.

However, there's one thing I want to mention: `comunity.vmware.vmware_guest_info` works, but `vmware.vmware.guest_info` failed with:

> Failed to connect to vCenter or ESXi API at ****:443 due to SSL verification failure : HTTPSConnectionPool(host='****', port=443): Max retries exceeded with url: /api (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)')))

It worked with `validate_certs: false` but this is something that might to be addressed. I'll test again and will provide more information, at least about the certificate of the vCenter.

mariolenz:
> But it's weird that I run into this SSL error. The certificate of the vCenter is signed by a private CA, but the OS trusts this CA. Both `curl` and `comunity.vmware.vmware_guest_info` work fine, but `vmware.vmware.guest_info` doesn't.

@mikemorency Looks like something connected to [requests](https://github.com/psf/requests) / [certifi](https://github.com/certifi/python-certifi). As far as I can see, it looks like they think "it's not a bug but a feature" to use their own set of trusted CAs. ATM I tend to disagree...

.github/workflows/ansible-test.yml
```diff
@@ -18,7 +18,10 @@ jobs:
         run: |
           set -euxo pipefail
           sudo apt-get update
-          sudo apt-get install podman
+
+          # recent version of podman has a bug. Fix is waiting deployment
```
bardielle, line 22: Is there a ticket in our kanban tracking it? 
If no, please open

plugins/module_utils/vmware.py
```diff
@@ -403,3 +404,22 @@ def get_resource_pool_by_name(self, pool_name, fail_on_missing=False):
         if not pool and fail_on_missing:
             self.module.fail_json("Unable to find resource pool with name %s" % pool_name)
         return pool
+
+    def get_all_objs_by_type(self, vimtype, folder=None, recurse=True):
```
bardielle, line 408:
```suggestion
    def list_objs_by_type(self, vimtype, folder=None, recurse=True):
```

bardielle, line 408: please add description

plugins/module_utils/vmware_facts.py
```diff
@@ -0,0 +1,430 @@
...
+        'hw_is_template': vm.config.template,
+        'hw_folder': None,
+        'hw_version': vm.config.version,
+        'instance_uuid': vm.config.instanceUuid,
+        'guest_tools_status': _get_vm_prop(vm, ('guest', 'toolsRunningStatus')),
+        'guest_tools_version': _get_vm_prop(vm, ('guest', 'toolsVersion')),
+        'guest_question': json.loads(json.dumps(vm.summary.runtime.question, cls=VmomiSupport.VmomiJSONEncoder,
+                                                sort_keys=True, strip_dynamic=True)),
+        'guest_consolidation_needed': vm.summary.runtime.consolidationNeeded,
+        'ipv4': None,
```
bardielle, line 62: Why does this value is None? (also ipv6)

plugins/module_utils/vmware_facts.py
```diff
@@ -0,0 +1,430 @@
...
+    for attribute in attributes:
+        try:
+            result = getattr(result, attribute)
+        except (AttributeError, IndexError):
+            return None
+    return result
+
+
+def gather_vm_facts(content, vm):
+    """ Gather facts from vim.VirtualMachine object. """
```
bardielle, line 37: This is long fucntion, please refactoring it into several function

machacekondra:
> > But it's weird that I run into this SSL error. The certificate of the vCenter is signed by a private CA, but the OS trusts this CA. Both `curl` and `comunity.vmware.vmware_guest_info` work fine, but `vmware.vmware.guest_info` doesn't.
> 
> @mikemorency Looks like something connected to [requests](https://github.com/psf/requests) / [certifi](https://github.com/certifi/python-certifi). As far as I can see, it looks like they think "it's not a bug but a feature" to use their own set of trusted CAs. ATM I tend to disagree...

I think it's worth to at least mention in documentation how to use system wide certificates, if don't support it by some parameter, but most of the users would be expect it to be used IMHO, they are not aware that we use `requests` library, so we should mention it in doc.

machacekondra (approved):

mikemorency:
The REST util is the only thing that relies on requests right now. I think that is why mario experienced the issue with this module and not the community one, it uses pyvmomi.

I added a note to the rest api docs. The user can set an environment variable to specify the ca bundle to use (instead of the certifi one provided).
REQUESTS_CA_BUNDLE, https://requests.readthedocs.io/en/latest/user/advanced/

mariolenz:
> I added a note to the rest api docs. The user can set an environment variable to specify the ca bundle to use (instead of the certifi one provided). REQUESTS_CA_BUNDLE, https://requests.readthedocs.io/en/latest/user/advanced/

Excellent! Thanks!

mariolenz (approved):
LGTM

bardielle (approved):

PR #52, opened 2024-07-22
Add checks in folder_template_from_vm tests

elsapassaro:
##### SUMMARY

Extending integration tests for folder_template_from_vm module:

- Create templates from a VM both in the same VM folder and in a different folder
- Add assertions to check that the newly created guest is indeed marked as template
- Use the module itself to remove the template

### DIFF
--- a/tests/integration/targets/vmware_folder_template_from_vm/tasks/main.yml
+++ b/tests/integration/targets/vmware_folder_template_from_vm/tasks/main.yml
@@ -46,6 +46,19 @@
       ansible.builtin.include_vars:
         file: ../group_vars.yml
 
+    - name: "Test setup: Create Folders"
+      community.vmware.vcenter_folder:
+        validate_certs: false
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        port: "{{ vcenter_port }}"
+        datacenter: "{{ vcenter_datacenter }}"
+        folder_name: "{{ item }}"
+        folder_type: vm
+        state: present
+      loop: "{{ template_folders }}"
+
     - name: "Test setup: Create VM guest {{ vm_name }}"
       community.vmware.vmware_guest:
         validate_certs: false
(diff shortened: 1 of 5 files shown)


tests/integration/targets/vmware_folder_template_from_vm/tasks/main.yml
```diff
@@ -98,7 +142,21 @@
...
-        folder: "{{ template_folder }}"
+        folder: "{{ item }}"
         state: absent
         force: true
         name: "{{ template_name }}"
+      loop:
+        - "{{ vm_folder }}"
+        - "{{ template_folder }}"
+
+    - name: "Test teardown: Delete Folder {{ template_folder }}"
```
anna-savina, line 153: what about `vm_folder`? Should it be deleted too?

elsapassaro: `vm_folder` is our `e2e-qe` folder which already exists and other tests are relying on its existence, so we don't delete it

elsapassaro: We might want to create the VM in a disposable folder instead. WDYT?

mikemorency (approved):

tests/integration/targets/vmware_folder_template_from_vm/tasks/main.yml
anna-savina: I think it's better create a new folder

shellymiron (approved):

tests/integration/targets/vmware_folder_template_from_vm/tasks/main.yml
elsapassaro: Done. 
BTW if we don't want to depend on the pre-existence of the `e2e-qe` VM folder then we should use the same approach in vmware_content_template integration tests [here](https://github.com/ansible-collections/vmware.vmware/blob/main/tests/integration/targets/vmware_content_template/defaults/main.yml#L6) and in the validated content tests as well

anna-savina: yes, you are right

prabinovRedhat (approved):

prabinovRedhat:
lgtm

mikemorency:
@elsapassaro  the test has failed twice now in the same spot but a re-run has fixed it. Im not sure if theres an issue or if its just the maintenance on the cluster messing things up?

elsapassaro:
@mikemorency I guess so, I also ran the playbook locally and it passed

PR #59, opened 2024-08-14
Add action_groups

mariolenz:
##### SUMMARY
I think we should implement an action group since this can be quite useful. I guess sooner or later someone will open a feature request for this, anyway.

amazon.aws and community.aws seem to share the action group `aws`. I suggest the same here: community.vmware already uses `vmware`, so let's also use it here.

Additionally, we might consider renaming the action group of the modules in vmware.vmware_rest from `vmware_rest` to `vmware`, but this is off-topic here. Just wanted to mention it as an idea.

If you think it would be better to not share action groups between those collections, that's also fine for me. The important thing is to have one in order to allow people to use module defaults. If you think it would be better to have a unique name for the action group, it would be great if you could suggest one.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
meta/runtime.yml

##### ADDITIONAL INFORMATION
ansible-collections/vmware.vmware_rest#440

### DIFF
--- a/changelogs/fragments/59-add-action_groups.yaml
+++ b/changelogs/fragments/59-add-action_groups.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Add action group (https://github.com/ansible-collections/vmware.vmware/pull/59).
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,5 +1,16 @@
 ---
 requires_ansible: '>=2.15.0'
+action_groups:
+  vmware:
+    - appliance_info
+    - cluster_drs
+    - content_template
+    - folder_template_from_vm
+    - guest_info
+    - license_info
+    - vcsa_settings
+    - vm_list_group_by_clusters_info
+    - vm_portgroup_info
 plugin_routing:
     modules:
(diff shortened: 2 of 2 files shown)


mikemorency:
im not familiar with action_groups. I read [this](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html#runtime-yml) but i still dont get it. What are they doing?

mariolenz:
The modules in this collection have a lot of common parameters / options like `hostname`, `username` and `password`. If you have several tasks using those modules you can specify that they're always the same. This a voids boilerplate / copy& paste because you can say "if you run a module from this group always use this parameters". I think you can overwrite them in any task, though.

[Module defaults groups](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_module_defaults.html#module-defaults-groups) mentions this as a feature preview, but I think it should be stable by now and the documentation is simply outdated.

mikemorency (approved):
that is pretty cool, i think its a good addition

mariolenz:
Thanks for reviewing @mikemorency!

bardielle:
@mariolenz thanks for your contribution

PR #61, opened 2024-08-21
migrate community.vmware.vmware_cluster_vcls module

mikemorency:
##### SUMMARY
This is a migration of the community.vmware.vmware_cluster_vcls module.

The inputs are backwards compatible. I added two inputs, `datastores_to_add` and `datastores_to_remove` to allow incremental modifications to the vCLS config. The legacy module only allowed absolute modifications; you needed to specify all allowed datastores every time.

The outputs that return any datastores added or removed by the module have changed key names to be more consistent with other output names in this collection. For example, `Added_AllowedDatastores` -> `added_datastores` 
Ive also added a new output to return the entire list of allowed datastores, `allowed_datastores`.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
cluster_vcl

##### Additional Information
Heres an example of the new output from the simulator tests:
```
{
    "added_datastores": [
        "LocalDS_0"
    ],
    "allowed_datastores": [
        "LocalDS_0"
    ],
    "changed": true,
    "invocation": {
        "module_args": {
            "allowed_datastores": [
                "LocalDS_0"
            ],
            "cluster": "DC0_C0",
            "datacenter": "DC0",
            "datastores_to_add": null,
            "datastores_to_remove": null,
            "hostname": "127.0.0.1",
            "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "port": 8989,
            "proxy_host": null,
            "proxy_port": null,
            "username": "user",
            "validate_certs": false
        }
    },
    "reconfig_task_result": {
        "completion_time": "2024-08-21T14:29:54.147149+00:00",
        "entity_name": "domain-c29",
        "error": null,
        "result": null,
        "state": "success"
    },
    "removed_datastores": []
}
```

### DIFF
--- a/tests/integration/targets/vmware_cluster_vcls/tasks/main.yml
+++ b/tests/integration/targets/vmware_cluster_vcls/tasks/main.yml
@@ -1,0 +1,117 @@
+---
+- name: Test On Simulator
+  when: run_on_simulator
+  block:
+    - name: Set vCLS Settings In Cluster
+      vmware.vmware.cluster_vcls:
+        validate_certs: false
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        datacenter: "{{ vcenter_datacenter }}"
+        cluster: "{{ test_cluster }}"
+        port: "{{ vcenter_port }}"
+        allowed_datastores: "{{ allowed_datastores }}"
+    # The simulator never seems to update its vCLS settings, so theres nothing to validate here
+
+- name: Test On VCenter
+  when: not run_on_simulator
+  block:
+    - name: Import common vars
+      ansible.builtin.include_vars:
+        file: ../group_vars.yml
+    - name: Create Test Cluster
+      vmware.vmware.cluster:
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        datacenter: "{{ vcenter_datacenter }}"
+        validate_certs: false
+        port: "{{ vcenter_port }}"
+        cluster_name: "{{ test_cluster }}"
+    - name: Set vCLS Allowed Datastores
+      vmware.vmware.cluster_vcls:
+        validate_certs: false
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        datacenter: "{{ vcenter_datacenter }}"
+        cluster: "{{ test_cluster }}"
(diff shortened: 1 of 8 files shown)


mariolenz (commented):
Do you _really_ want to migrate this module? [vSphere 8.0 Update 3 introduces a redesign of vCLS to Embedded vCLS, which utilizes vSphere Pod technology](https://docs.vmware.com/en/VMware-vSphere/8.0/rn/vsphere-esxi-803-release-notes/index.html).

This means that [embedded vCLS VMs (introduced in 8.0 U3) have no datastores](https://knowledge.broadcom.com/external/article/370225/scripting-impacts-of-embedded-vcls-vms-h.html). So you would migrate a module that looks to be quite useless for vSphere 8.0 U3+.

I suggest to _not_ migrate this module. You can do it, if you really want to... but personally I wouldn't. Doesn't make much sense to me. Let's keep it in `community.vmware`.

I'm not against migrating the module, but since I don't see a future for it...

mikemorency:
We have a downstream repo that uses this module, and I think for now we still need to support people using 7x. The additional testing we can do in this repo would be helpful, although i guess once we update our vcenter the tests there will be less helpful....

@machacekondra what do you think

mariolenz:
> We have a downstream repo that uses this module, and I think for now we still need to support people using 7x.

No problem, just wanted to make you aware that you're migrating a module that won't be useful for long.

I'm not sure when 7.x will go EOL. It used to be April 2025 or something like this, but I think I've read somewhere that this has been extended some 6 months or so. Can't find a link for this ATM, though :-/

> The additional testing we can do in this repo would be helpful

True.

> although i guess once we update our vcenter the tests there will be less helpful...

Also true :-)

tests/integration/targets/vmware_cluster_vcls/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+    # The simulator never seems to update its vCLS settings, so theres nothing to validate here
+
+- name: Test On VCenter
+  when: not run_on_simulator
+  block:
+    - name: Import common vars
+      ansible.builtin.include_vars:
+        file: ../group_vars.yml
+    - name: Create Test Cluster
+      community.vmware.vmware_cluster:
```
elsapassaro, line 24: wouldn't it be better to use `vmware.vmware.cluster` module instead now that it's merged? 

same for the task to destroy the test cluster

mariolenz (changes requested):
Please add the new `cluster_vcls` module to the `vmware` action group in `meta/runtime.yml`.

anna-savina (approved):
LGTM

PR #60, opened 2024-08-20
migrate community.vmware.vmware_cluster module

mikemorency:
##### SUMMARY
This is a migration of the community.vmware.vmware_cluster module.

The inputs have remained the same (although more aliases are allowed for the `cluster` attribute).

Ive added identifying information about the cluster in the outputs. When a cluster is destroyed, only the name is returned. When a cluster is created (or already exists) the cluster name and moid are returned.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
cluster

PR #63, opened 2024-08-26
Add cluster to action group

mariolenz:
##### SUMMARY
This is a follow-up to both #59 and #60 to add the `cluster` module to the `action_group`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
cluster
meta/runtime.yml

##### ADDITIONAL INFORMATION
Since there hasn't been a release containing this module yet, I made this a trivial change. I don't think this needs to show up in the changelog. Let me know if you think otherwise and I'll change it.

### DIFF
--- a/changelogs/fragments/63-add-cluster-module-to-action_groups.yaml
+++ b/changelogs/fragments/63-add-cluster-module-to-action_groups.yaml
@@ -1,0 +1,3 @@
+---
+trivial:
+  - cluster - Add cluster module to action group (https://github.com/ansible-collections/vmware.vmware/pull/63).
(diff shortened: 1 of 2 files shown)


mariolenz:
Any idea why `Ansible Eco vCenter Integration Test / ansible_integration_test (pull_request_target)` fails? This doesn't look like it can be caused by the changes in my PR. I don't understand this :shrug:

mikemorency (approved):

mikemorency:
the portgroup failures were transient, but i found the another test was failing like 80% of the time. seems like there was a bug in the module
https://github.com/ansible-collections/vmware.vmware/pull/64

PR #64, opened 2024-08-28
fix integration test failures for folder_template_from_vm

mikemorency:
##### SUMMARY
The `folder_template_from_vm` plugin was one of the first to be migrated from the community collection. It included a very simple method to wait for a running task in vmware. It looks like that method did not correctly evaluate the state of running tasks. This meant that running tasks would be marked as failures.

We now have a much better utility for waiting for running tasks. This change makes the plugin use that instead.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
folder_template_from_vm


mariolenz (approved):
LGTM

Thanks @mikemorency!

PR #70, opened 2024-09-26
migrate cluster_drs_recommendations module

mikemorency:
##### SUMMARY
This migrates the cluster_drs_recommendations module from the community.vmware collection

The inputs have not changed.
The outputs have changed from the original module. This version reports a description of the applied recommendations as well as details about the vcenter task associated with the change

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
cluster_drs_recommendations

### DIFF
--- a/plugins/modules/cluster_drs_recommendations.py
+++ b/plugins/modules/cluster_drs_recommendations.py
@@ -1,0 +1,190 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2023, Ansible Cloud Team (@ansible-collections)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: cluster_drs_recommendations
+short_description: Apply Distributed Resource Scheduler (DRS) recommendations on VMware vSphere clusters
+description:
+    - Applies DRS recommendations on VMware vSphere clusters.
+author:
+    - Ansible Cloud Team (@ansible-collections)
+
+options:
+    cluster:
+        description:
+            - The name of the cluster to be managed.
+        type: str
+        required: true
+        aliases: [ cluster_name ]
+    datacenter:
+        description:
+            - The name of the datacenter.
+        type: str
+        required: true
+        aliases: [ datacenter_name ]
+
+extends_documentation_fragment:
+    - vmware.vmware.vmware.documentation
+'''
+
+EXAMPLES = r'''
+- name: Apply DRS Recommendations for Cluster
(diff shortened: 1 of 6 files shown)


anna-savina:
@mikemorency what do you think about next questions?

1. Is there an alternative approach to testing this module? 
Currently, the test cluster is empty, and there are no recommendations, meaning the test doesn't actually perform anything.

2. What should the result be if there are no recommendations? 
The `community.vmware.vmware_cluster_drs_recommendations` module records in the results that no changes occurred
```
    "recommendations_info": {
        "changed": false,
        "failed": false,
        "result": "No recommendations."
    }
```

while `vmware.vmware.cluster_drs_recommendations` shows that changes were made
```
    "recommendations_info": {
        "applied_recommendations": [],
        "changed": true,
        "failed": false
    }
```

mikemorency:
@anna-savina 
1. Validating that recommendations are made and run requires a vcenter with multiple hosts running multiple VMs in an unbalanced way. I think setting up that testing scenario will take awhile. I think that might be too long for the PR tests we run, what do you think?
I was going to mention this in our QE sync but maybe we could have a longer test that runs nightly or something. That could stand up an entire test vcenter and then fully test this and other modules. Or maybe we could create some test VM templates (esxi, rhel9, etc) so spinning up a new complete cluster is much faster?

2. Your right, the module should not report that something has changed if theres no recommendations. Ill fix that

anna-savina:
> I think setting up that testing scenario will take awhile. I think that might be too long for the PR tests we run, what do you think?

yes, I agree with you. I didn't find right approach to test this module, thought maybe you have ideas.

> I was going to mention this in our QE sync but maybe we could have a longer test that runs nightly or something.

I'm not sure about this. Let's discuss it during our QE sync meeting

bardielle (approved):

plugins/modules/cluster_drs_recommendations.py
```diff
@@ -0,0 +1,191 @@
...
+
+EXAMPLES = r'''
+- name: Apply DRS Recommendations for Cluster
+  vmware.vmware.cluster_drs_recommendations:
+    hostname: '{{ vcenter_hostname }}'
+    username: '{{ vcenter_username }}'
+    password: '{{ vcenter_password }}'
+    datacenter_name: datacenter
+    cluster_name: cluster
+    enable: true
```
anna-savina, line 47: Is 'enable' supported parameter?
`FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/bin/python3"}, "changed": false, "msg": "Unsupported parameters for (vmware.vmware.cluster_drs_recommendations) module: enable. Supported parameters include: cluster, datacenter, hostname, password, port, proxy_host, proxy_port, username, validate_certs (admin, cluster_name, datacenter_name, pass, pwd, user)."}`

mikemorency: Nope, i missed that in the example docs. Ill remove it. I dont see it anywhere else (plugin code, tests, etc), let me know if you do

anna-savina (approved):
LGTM

Issue #68, opened 2024-09-19
vmware.vmware.cluster_drs has drs_vmotion_rate inverted


MallocArray:
##### SUMMARY

`vmware.vmware.cluster_drs` documentation shows that drs_vmotion_rate is 1 (lowest) to 5 (highest), but setting value 2 results in vCenter 8.0 U3 setting the DRS Migration Threshold to 4 which is toward the Aggressive (More Frequent vMotions) side of the slider.

The previous `community.vmware.vmware_cluster_drs` module handled this as expected, with a value of 2 resulting in a Migration Threshold of 2 which is toward the Conservative end of the slider.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

vmware.vmware.cluster_drs

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.11]
  config file = /runner/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.11/site-packages/ansible
  ansible collection location = /runner/collections
  executable location = /usr/local/bin/ansible
  python version = 3.11.9 (main, Aug 23 2024, 00:00:00) [GCC 11.5.0 20240719 (Red Hat 11.5.0-2)] (/usr/bin/python3.11)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection    Version
------------- -------
vmware.vmware 1.5.0  
```

##### CONFIGURATION

```paste below
COLLECTIONS_PATHS(/runner/ansible.cfg) = ['/runner/collections']
CONFIG_FILE() = /runner/ansible.cfg
DEFAULT_FILTER_PLUGIN_PATH(/runner/ansible.cfg) = ['/runner/custom_filters']
DEFAULT_ROLES_PATH(/runner/ansible.cfg) = ['/runner/roles']
DEFAULT_STRATEGY_PLUGIN_PATH(/runner/ansible.cfg) = ['/runner/custom_plugins/mitogen-0.3.4/ansible_mitogen/plugins/strategy']
DEFAULT_TIMEOUT(/runner/ansible.cfg) = 40
HOST_KEY_CHECKING(/runner/ansible.cfg) = False
PARAMIKO_LOOK_FOR_KEYS(/runner/ansible.cfg) = False
```

##### OS / ENVIRONMENT

Execution Environment based on Stream9

[...]

PR #73, opened 2024-09-30
Mm bugfix/drs threshold backwards

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/68

When setting the DRS vmotion rate, the user input value gets reversed by the vCenter API (so 1 becomes 5 and 5 becomes 1). This is confusing because the UI and documentation do not indicate this in anyway.
This fix makes the module reverse the numbers in code, so if a user chooses a rate of 1, we change it to 5 in code before inputting it in the API. It will display as 1 in the UI which is what the user wanted in the first place.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cluster_drs

##### ADDITIONAL INFORMATION
The `community.vmware.vcenter_cluster_info` module reports back what the API says. This means that if the user sets the DRS rate to 2, the cluster info module will say the DRS rate is 4. Obviously this is confusing, so when the info module is migrated to this repo, we should fix that as well. I left a note in our tests (which use that module) and made a jira ticket to follow up.

### DIFF
--- a/plugins/modules/cluster_drs.py
+++ b/plugins/modules/cluster_drs.py
@@ -158,7 +158,6 @@ class VMwareCluster(PyVmomi):
         self.cluster = self.get_cluster_by_name(self.params.get('cluster'), fail_on_missing=True, datacenter=datacenter)
 
         self.enable_drs = self.params.get('enable')
-        self.drs_vmotion_rate = self.params.get('drs_vmotion_rate')
         self.drs_enable_vm_behavior_overrides = self.params.get('drs_enable_vm_behavior_overrides')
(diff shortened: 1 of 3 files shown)


plugins/modules/cluster_drs.py
```diff
@@ -168,6 +167,10 @@ def __init__(self, module):
             self.cluster.configurationEx.drsConfig.option
         )
 
+    @property
+    def drs_vmotion_rate(self):
```
bardielle, line 171: Please add a comment or explanation for that rate value

bardielle (approved):

PR #75, opened 2024-10-08
fix dpm automation level option

mikemorency:
##### SUMMARY
The automation level option for dpm_cluster did not match the options that are used in the UI. This is because the API uses a different word (automated vs automatic).

This change modifies the module options to match what the user sees in the vcenter UI and docs.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cluster_dpm

### DIFF
--- a/changelogs/fragments/75-fix-dpm-automation-level-terms.yml
+++ b/changelogs/fragments/75-fix-dpm-automation-level-terms.yml
@@ -1,0 +1,3 @@
+---
+minor_change:
+  - cluster_dpm - Change automation level option from 'automated' to 'automatic' so it matches the vCenter UI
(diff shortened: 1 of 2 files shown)


prabinovRedhat (approved):
lgtm

mariolenz (commented):
It looks like the module hasn't been released yet, right? In this case, I think it's OK.

Otherwise, I should say it's a **breaking change** which requires a new major version. But if the module really hasn't been released yet, we can't break any existing playbooks :-)

mikemorency:
yea, i just merged this module a few days ago so it hasnt been released yet

mariolenz (approved):
LGTM

PR #81, opened 2024-11-07
Mm feature/migrate content library item info

mikemorency:
##### SUMMARY
This is migrating the content_library_item_info from the vmware_rest collection to here. The auto-generated module in the rest collection does not work due to an issue with the OpenAPI spec, so its easier to add the functionality to this collection instead.

I also consolidated some of the content library shared methods in the rest api module_util

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
content_library_item_info

### DIFF
--- a/plugins/module_utils/_vmware_rest_client.py
+++ b/plugins/module_utils/_vmware_rest_client.py
@@ -26,7 +26,6 @@ try:
     from com.vmware.vapi.std_client import DynamicID
     from vmware.vapi.vsphere.client import create_vsphere_client
     from com.vmware.vapi.std.errors_client import Unauthorized
-    from com.vmware.content.library_client import Item
     from com.vmware.vcenter_client import (Folder,
                                            Datacenter,
                                            ResourcePool,
@@ -63,6 +62,9 @@ class VmwareRestClient(object):
         self.params = module.params
         self.check_required_library()
         self.api_client = self.connect_to_vsphere_client()
+
+        self.library_service = self.api_client.content.Library
+        self.library_item_service = self.api_client.content.library.Item
 
     # Helper function
     def get_error_message(self, error):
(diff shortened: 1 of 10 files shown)


tests/integration/targets/vmware_content_library_item_info/vars.yml
```diff
@@ -0,0 +1,6 @@
+vcenter_hostname: "127.0.0.1"
```
anna-savina, line 1: I don't see test for simulator. Will you add it later?

tests/integration/targets/vmware_content_library_item_info/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+          - (__res.library_item_info | length) > 0
+
+    - name: Gather Fake Library Info
+      vmware.vmware.content_library_item_info:
+        validate_certs: false
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        port: "{{ vcenter_port }}"
+        library_item_name: should_not_exist
```
anna-savina, line 32: Should be `library_name` specified?

tests/integration/targets/vmware_content_library_item_info/vars.yml
mikemorency: I cannot figure out how to mock the rest api for this module, so I dont think I can include the simulator tests at this time

anna-savina: In this case better delete `vars.yml` and remove it from `run.yml`

anna-savina (approved):
LGTM

anna-savina (approved):

plugins/module_utils/_vmware_rest_client.py
```diff
@@ -191,51 +193,54 @@ def get_vm_by_name(self, name):
 
         return vms[0]
 
-    def get_library_item_by_name(self, name):
+    def get_content_library_ids(self, name=None, library_type=None, fail_on_missing=False):
```
bardielle, line 196: Why did you change this behavior?

bardielle: I saw that the other functions are still by name
@machacekondra what do you think?

mariolenz: @machacekondra @mikemorency I don't think so because they're all marked as private. If people use them and they break backward compatibility, that's their problem. That's what private means.

Apropos of nothing, it would be great if there would be some non-private module utils that I could re-use in `community.vmware`. Actually, I think this was the original plan :thinking:

mikemorency: Yes we did make them private while we were still figuring out method naming and making changes. Sounds like we can stick with whats in the PR.

@mariolenz , Most of the pyvmomi/rest api base utils stay the same, but I still make the occasional functional change to the other utils. Ill make a ticket to review the function names and prep the pyvmomi/rest utils to make them public. Ill have to look at previous conversation to see if that can be done in a minor release or needs a major one

PR #85, opened 2024-12-03
Feature/prep module utils

mikemorency:
##### SUMMARY
This is preparing the pymomi module utils to be made public. Most of the method names were updated in pyvmomi to reflect what they actually do. Additionally some of the documentation fragments and arg specs can be consolidated, as there is no significant difference between the fragments we do have.

I am not able to update the rest utils yet. I think i see some possible bugs with the current methods and I want to confirm and fix those first

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils/_vmware

### DIFF
--- a/plugins/doc_fragments/vmware.py
+++ b/plugins/doc_fragments/vmware.py
@@ -13,119 +13,62 @@ __metaclass__ = type
 class ModuleDocFragment(object):
     # Parameters for VMware modules
     DOCUMENTATION = r'''
-notes:
-  - All modules require API write access and hence is not supported on a free ESXi license.
-  - All variables and VMware object names are case sensitive.
-options:
-    hostname:
-      description:
-      - The hostname or IP address of the vSphere vCenter or ESXi server.
-      - If the value is not specified in the task, the value of environment variable E(VMWARE_HOST) will be used instead.
-      type: str
-    username:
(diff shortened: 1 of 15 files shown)


mikemorency (commented):
i think we could also remove the `vmware` prefix from the module_utils file names. I dont really see a need for it. Any opinions?

mariolenz (commented):
> This is preparing the pymomi module utils to be made public.

Thanks, I really appreciate this!

But it looks like there are quite a lot of changes. I feel a little bit overwhelmed. Would it be possible to break this down to several PRs? For example, it looks like you've just added some indentation to `plugins/doc_fragments/vmware.py`. Would it be possible to have a PR just for this?

There are some other changes that I'm not too sure about. But it's hard to discuss them. This PR is quite a mouthful (at least to me).

> i think we could also remove the `vmware` prefix from the module_utils file names. I dont really see a need for it. Any opinions?

I fully agree. `vmware.vmware.plugins.module_utils._vmware` is a little bit redundant. I see the same thing in `comunity.vmware` where most modules start with `vmware_`, too. But that's my problem, not yours ;-)

Anyway, I think this also would be easier to fix / discuss if we have a separate PR just on this.

PR #86, opened 2024-12-04
Feature/centralize doc fragments

mikemorency:
##### SUMMARY
This is preparing the pymomi module utils to be made public. The main change here is cleaning up the doc fragments. Eventually, we can combine the rest and pyvmomi doc fragments. The only significant differences between them is
1. pyvmomi has two versions of the doc frags, one with cluster/datacenter and one without
2. rest has the `protocol` parameter

So far in this repo, the cluster/datacenter documentation is re-written in every module so its not needed in the doc frag. I think thats a good trend to carry on going forward, and it has the added benefit of making the writer consider if cluster and datacenter are required, optional, or unused in their module.

The protocol parameter is unused by pyvmomi, but I dont think theres any harm in including it. Iit makes our lives easier since we dont need to maintain two copies of the doc frag thats 95% redundant. And theres no harm in the user setting it for pyvmomi since that SDK automatically determines the protocol anyway 

This change does not update the REST utils doc fragments, that should come later

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils/_vmware doc fragments


mikemorency:
FYI this will need to be combined with https://github.com/ansible-collections/vmware.vmware/pull/87 before it can be merged, but I am going to leave them separate for now so its easier to review

plugins/doc_fragments/vmware.py
mikemorency, line 1: this is a frusterating diff from github. 
1. added the note about the requests library
2. added the `proxy_protocol` option with an alias to `protocol`
3. removed the VCENTER_DOCUMENTATION block

mariolenz:
https://github.com/ansible-collections/vmware.vmware/pull/87#issuecomment-2525191381

mikemorency:
closing in favor of https://github.com/ansible-collections/vmware.vmware/pull/89

PR #87, opened 2024-12-04
centralize argument spec

mikemorency:
##### SUMMARY
This is preparing the pymomi module utils to be made public. The main change is centralizing the argument spec. This is related to the doc fragment update in https://github.com/ansible-collections/vmware.vmware/pull/86

pyvmomi and rest both have 95% of the same args and the differences can be resolved without impacting the end user. In this change, I removed the cluster/datacenter from the common arg spec since they are re-defined in all of the modules that need them anyway. I also added the protocol arg to the spec. Since its used by the rest api but does nothing for pymomi, we can safely add it to all modules and not need to maintain two copies of this arg spec

Rest API modules still use their own arg spec, but they should be updated to use this one later on

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils/_vmware

### DIFF
--- a/changelogs/fragments/87-centralize-spec-args.yml
+++ b/changelogs/fragments/87-centralize-spec-args.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - pyvmomi doc fragments - Remove redundant 'vcenter_documentation' fragment. Add proxy protocol documentation to all modules for simplicity. Remove redundant cluster and datastore docs
--- a/plugins/module_utils/_vmware.py
+++ b/plugins/module_utils/_vmware.py
@@ -32,52 +32,12 @@ except ImportError:
     PYVMOMI_IMP_ERR = traceback.format_exc()
     HAS_PYVMOMI = False
 
-from ansible.module_utils.basic import env_fallback, missing_required_lib
+from ansible.module_utils.basic import missing_required_lib
 
 
 class ApiAccessError(Exception):
     def __init__(self, *args, **kwargs):
         super(ApiAccessError, self).__init__(*args, **kwargs)
-
-
-def vmware_argument_spec():
-    return dict(
-        hostname=dict(type='str',
-                      required=False,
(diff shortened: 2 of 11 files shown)


mikemorency:
FYI, this will need to be combined with https://github.com/ansible-collections/vmware.vmware/pull/86 before merging, but Im going to leave it as is for now so its easier to review

mariolenz:
> I also added the protocol arg to the spec. Since its used by the rest api but does nothing for pymomi, we can safely add it to all modules

I'm not sure that it's safe to add this to all modules. If we do, we have to document this. And if it's documented, people will assume it _works_ for those modules. You're right, there should be no change in functionality. But it might make users expect things that are, at least ATM, not implemented ;-)

> and not need to maintain two copies of this arg spec

I'm with you there. But instead of adding an argument to the pyvmomi modules that they don't use, how about dropping it from common_argument_spec() and have another function that simply re-uses it? Something like:

```
def vmware_rest_argument_spec():
    return {
        **common_argument_spec(), **dict(proxy_protocol=dict(
            proxy_protocol=dict(
                type='str',
                default='https',
                choices=['https', 'http'],
                aliases=['protocol']
            )
```

I've made this up quick'n'dirty, not sure if it works or if I made any stupid mistakes. But I hope this makes clear what I mean. I guess we could also do something similar for the docs fragments (#86).

mikemorency:
fair enough. I did what you suggested for the arg spec and I think its works well. For the doc frags I did something similar. 

Let me know your thoughts, I can try combining the two PRs if you think it looks good and well see what the build says

mariolenz:
> Let me know your thoughts, I can try combining the two PRs if you think it looks good and well see what the build says

I'm not really sure ATM. Maybe you're right and it would be a good idea to combine both PRs (again?) and see what happens :shrug:

mikemorency:
closing in favor of https://github.com/ansible-collections/vmware.vmware/pull/89

PR #88, opened 2024-12-04
Feature/standardize pyvmomi method names

mikemorency:
##### SUMMARY
This is preparing the pymomi module utils to be made public. There are two main changes in this PR, all relating to module_utils/_vmware

1. Standardize the get_<something>_by_name_or_moid methods. These methods are all very similar but some slightly differ just because they were written at different times. This change updates the names to be both descriptive and similar to one another, updates documentation for the methods to be similar to one another, and adds an optional parameter to throw a standard error if the search comes back empty

2. `list_all_objs_by_type` doesnt match the get_ method name pattern we have and needlessly return a dict. Updated the name to be `get_all....` and changed the return value to a list, like the other methods

I am not able to update the rest utils yet. I think i see some possible bugs with the current methods and I want to confirm and fix those first

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils/_vmware

### DIFF
--- a/plugins/module_utils/_vmware.py
+++ b/plugins/module_utils/_vmware.py
@@ -262,11 +262,12 @@ class PyVmomi(object):
             # for backwards-compat
             return None
 
-    def get_dvs_portgroup_by_name_or_moid(self, identifier, fail_on_missing=False):
+    def get_standard_portgroup_by_name_or_moid(self, identifier, fail_on_missing=False):
         """
         Get a portgroup from type 'STANDARD_PORTGROUP' based on name or MOID
         Args:
             identifier: The name or the ID of the portgroup
+            fail_on_missing: If true, an error will be thrown if no networks are found
         Returns:
             The standard portgroup object
         """
@@ -280,6 +281,7 @@ class PyVmomi(object):
         Get a portgroup from type 'DISTRIBUTED_PORTGROUP' based on name or MOID
         Args:
             identifier: The name or the ID of the portgroup
+            fail_on_missing: If true, an error will be thrown if no networks are found
         Returns:
             The distributed portgroup object
         """
@@ -344,7 +346,7 @@ class PyVmomi(object):
 
     def get_folders_by_name_or_moid(self, identifier, fail_on_missing=False):
         """
-            Get all folders with the given name. Names are not unique
+            Get all folders with the given name or MOID. Names are not unique
             in a given cluster, so multiple folder objects can be returned
             Args:
                 identifier: Name or MOID of the folder to search for
@@ -376,7 +378,7 @@ class PyVmomi(object):
 
     def get_datastore_by_name_or_moid(self, identifier, fail_on_missing=False):
         """
-            Get the datastore matching the given name. Datastore names must be unique
+            Get the datastore matching the given name or MOID. Datastore names must be unique
             in a given cluster, so only one object is returned at most.
             Args:
                 identifier: Name or MOID of the datastore to search for
@@ -391,7 +393,7 @@ class PyVmomi(object):
 
     def get_resource_pool_by_name_or_moid(self, identifier, fail_on_missing=False):
         """
-            Get the resource pool matching the given name. Pool names must be unique
+            Get the resource pool matching the given name or MOID. Pool names must be unique
             in a given cluster, so only one object is returned at most.
             Args:
                 identifier: Name or MOID of the pool to search for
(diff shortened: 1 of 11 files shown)


plugins/module_utils/_vmware.py
```diff
@@ -262,25 +262,31 @@ def get_objs_by_name_or_moid(self, vimtype, name, return_all=False, search_root_
...
+        return pg
 
-    def get_dvs_portgroup(self, portgroup):
+    def get_dvs_portgroup_by_name_or_moid(self, identifier, fail_on_missing=False):
         """
-        Get a portgroup from type 'DISTRIBUTED_PORTGROUP'
+        Get a portgroup from type 'DISTRIBUTED_PORTGROUP' based on name or MOID
         Args:
-            portgroup: The name or the ID of the portgroup
+            identifier: The name or the ID of the portgroup
```
anna-savina, line 282:
```suggestion
            identifier: The name or the ID of the portgroup
            fail_on_missing: If true, an error will be thrown if no portgroup is found
```

plugins/module_utils/_vmware.py
```diff
@@ -262,25 +262,31 @@ def get_objs_by_name_or_moid(self, vimtype, name, return_all=False, search_root_
...
             return None
 
-    def get_standard_portgroup(self, portgroup):
+    def get_dvs_portgroup_by_name_or_moid(self, identifier, fail_on_missing=False):
         """
-        Get a portgroup from type 'STANDARD_PORTGROUP'
+        Get a portgroup from type 'STANDARD_PORTGROUP' based on name or MOID
         Args:
-            portgroup: The name or the ID of the portgroup
+            identifier: The name or the ID of the portgroup
```
anna-savina, line 269:
```suggestion
            identifier: The name or the ID of the portgroup
            fail_on_missing: If true, an error will be thrown if no portgroup is found
```

plugins/module_utils/_vmware.py
```diff
@@ -368,88 +374,94 @@ def get_folder_by_absolute_path(self, folder_path, fail_on_missing=False):
...
+        ds = self.get_objs_by_name_or_moid([vim.Datacenter], identifier)
         if not ds and fail_on_missing:
-            self.module.fail_json("Unable to find datacenter with name %s" % dc_name)
+            self.module.fail_json("Unable to find datacenter with name or MOID %s" % identifier)
         return ds
 
-    def get_cluster_by_name(self, cluster_name, fail_on_missing=False, datacenter=None):
+    def get_cluster_by_name_or_moid(self, identifier, fail_on_missing=False, datacenter=None):
         """
             Get the cluster matching the given name. Cluster names must be unique
```
anna-savina, line 461:
```suggestion
            Get the cluster matching the given name or moid. Cluster names must be unique
```

plugins/module_utils/_vmware.py
```diff
@@ -368,88 +374,94 @@ def get_folder_by_absolute_path(self, folder_path, fail_on_missing=False):
...
             in a given vcenter, so only one object is returned at most.
             Args:
-                cluster_name: Name of the cluster to search for
+                identifier: Name or MOID of the cluster to search for
                 fail_on_missing: If true, an error will be thrown if no clusters are found
```
anna-savina, line 465:
```suggestion
                fail_on_missing: If true, an error will be thrown if no clusters are found
                datacenter: Name of the datacenter to search for. If not provided, all datacenters will be examined
```

plugins/module_utils/_vmware.py
```diff
@@ -262,25 +262,31 @@ def get_objs_by_name_or_moid(self, vimtype, name, return_all=False, search_root_
             # for backwards-compat
             return None
 
-    def get_standard_portgroup(self, portgroup):
+    def get_dvs_portgroup_by_name_or_moid(self, identifier, fail_on_missing=False):
```
machacekondra, line 265: I think this is wrong name of the method. shouldn't be `get_standard_portgroup_by_name_or_moid`?

mikemorency: ahh good catch. interesting that none of the tests failed since its a duplicate method name....

plugins/module_utils/_vmware.py
mikemorency: this was missing in a few other places, i added it there as well

machacekondra (approved):

PR #91, opened 2024-12-16
Feature/migrate esxi host inventory

mikemorency:
##### SUMMARY
Adds an ESXi host inventory plugin. This plugin was originally going to be a migration from the community collection, but I dont think I can call it that at this point. 

While the output is largely the same, some redundant properties have been removed and some data has been manipulated to be more useful (i.e. tags now return the tag ID and tag name instead of just the tag name, which is not unique).
Additionally, some community options seemed confusing, conflicted with options we have for info modules, or are no longer relevant. 

Also, Im not able to use the existing pyvmomi class in the inventory plugin since it is tightly coupled to modules. The community collection works around this by maintaining two nearly identical versions of the pyvmomi/rest utilities for modules and for plugins. 
I think we can break some of the pyvmomi/rest utils into more generic "client" classes. Those can be used by module or plugin base classes (like i do in this PR for the inventory plugin), and then those base classes can be extended by individual modules and plugins. 
This approach has greatly reduced the amount of duplicated code, and I think will make it easier for community to switch to the new utils.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
vmware.vmware.esxi_hosts

##### ADDITIONAL INFO
I tested the new "client" classes with the modules and it worked as expected. Fully switching them over would be a task for next release

### DIFF
--- a/plugins/doc_fragments/plugin_base_options.py
+++ b/plugins/doc_fragments/plugin_base_options.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016, Charles Paul <cpaul@ansible.com>
+# Copyright: (c) 2018, Ansible Project
+# Copyright: (c) 2019, Abhijeet Kasurde <akasurde@redhat.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+    # This document fragment serves as a partial base for all vmware plugins. It should be used in addition to the base fragment, vmware.vmware.base_options
+    # since that contains the actual argument descriptions and defaults. This just defines the environment variables since plugins have something
+    # like the module spec where that is usually done.
+    DOCUMENTATION = r'''
+options:
+  hostname:
+    env:
+      - name: VMWARE_HOST
+  username:
+    env:
+      - name: VMWARE_USER
+  password:
+    env:
+      - name: VMWARE_PASSWORD
+  validate_certs:
+    env:
+      - name: VMWARE_VALIDATE_CERTS
+  port:
+    env:
+      - name: VMWARE_PORT
+  proxy_host:
+    env:
+      - name: VMWARE_PROXY_HOST
+  proxy_port:
+    env:
+      - name: VMWARE_PROXY_PORT
+'''
--- a/tests/integration/targets/vmware_inventory_esxi_hosts/tasks/main.yml
+++ b/tests/integration/targets/vmware_inventory_esxi_hosts/tasks/main.yml
@@ -1,0 +1,51 @@
+---
+- block:
+    - name: Import common vars
+      ansible.builtin.include_vars:
+        file: ../group_vars.yml
+
+    # the ansible-inventory process does not have access to any of the variables in this playbook,
+    # so the auth vars are exposed as env vars
+    - name: Run Inventory Plugin
+      ansible.builtin.command: ansible-inventory -i "{{ role_path }}/files/test.esxi_hosts.yml" --list
+      register: _inventory_out
+      environment:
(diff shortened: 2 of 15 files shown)


tests/integration/targets/vmware_inventory_esxi_hosts/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+      ansible.builtin.set_fact:
+        inventory_results: "{{ _inventory_out.stdout | from_json }}"
+
+    # you can't reference the 'all' property here for some reason. It reverts back to the test playbook inventory
+    # instead of the inventory_results
+    - name: Check Output
+      ansible.builtin.assert:
+        that:
+          - first_host.ansible_host
+          - first_host.tags is defined
```
prabinovRedhat, line 21:
```suggestion
          - first_host.tags is defined and first_host.tags is not none
```
If tags is defined but has a None value, tags is defined will still pass.
To catch a None value explicitly, might be better to add and tags is not none ?

prabinovRedhat: as well for ansible_host

prabinovRedhat:
I see that the test failed due to the following error:

`The conditional check 'first_host.ansible_host' failed. The error was: error while evaluating conditional (first_host.ansible_host): {{ (inventory_results._meta.hostvars.values() | first) }}: No first item, sequence was empty.`

So, the inventory plugin did not return any hosts, causing inventory_results._meta.hostvars to be empty

plugins/doc_fragments/plugin_base_options.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016, Charles Paul <cpaul@ansible.com>
+# Copyright: (c) 2018, Ansible Project
+# Copyright: (c) 2019, Abhijeet Kasurde <akasurde@redhat.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
```
shellymiron, line 7: check

machacekondra (approved):

tests/integration/targets/vmware_inventory_esxi_hosts/tasks/main.yml
mikemorency: thats a good idea. For some reason I wasnt able to test for None, but I am able to test if the variable is a dictionary. So I added that. I also added a check to make sure ansible_host is an IP

prabinovRedhat: Great, thanks

prabinovRedhat: I just noticed the regex (^[\d+\.]+$) only works for IPv4 and doesn’t handle IPv6. Should we assume first_host.ansible_host will always be IPv4, or do we need to update the regex to support IPv6 too?

mikemorency: we should check for ipv6 too!

prabinovRedhat: ok :) so, just fix that and I think the PR is ok
thank you

prabinovRedhat: `- first_host.ansible_host is regex('^((\\d{1,3}\\.){3}\\d{1,3}|([0-9a-fA-F:]+))$')`

smth like that?

prabinovRedhat:
Screenshot 2024-12-19 at 13 50 14

While testing the inventory plugin, I noticed that the initial output format looks a bit messy and difficult to read (as shown in the attached screenshot). However, below, I do see a more structured and readable version.

Could you confirm if this initial "user unfriendly" format is expected behavior? If so, is it something that could potentially be optimized for better readability during debugging?

mikemorency:
@prabinovRedhat 
I can try to print some more readable output first and then show the full output if the pretty print fails

prabinovRedhat:
Thanks looks good!

PR #92, opened 2024-12-19
fix content_template library_id reference in delete

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/90

While refactoring the content_template module, one variable reference was missed. There are no tests for the spot that this issue occurred. I fixed the variable and added tests to cover this spot.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
content_template

PR #93, opened 2024-12-23
vm_name now not a required variable

OrrBG:
##### SUMMARY
Fixes: vm_name now not a required variable

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
content_template module


prabinovRedhat:
lgtm

prabinovRedhat (approved):

mikemorency (changes requested):
`vm_name` is still required when state==present. The module needs to check for that, preferably in the argument spec definition. And the docs need to reflect that

anna-savina:
re-base branch please. You branch doesn't include changes from [92](https://github.com/ansible-collections/vmware.vmware/pull/92)

plugins/modules/content_template.py
```diff
@@ -206,7 +206,7 @@ def main():
     argument_spec.update(
         template=dict(type='str', required=True),
         library=dict(type='str', required=True),
-        vm_name=dict(type='str', required=True),
+        vm_name=dict(type='str', required=(argument_spec['state'] == 'present')),
```
mikemorency, line 209: this is not the correct syntax, see `required-if` here 
https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec-dependencies

plugins/modules/content_template.py
```diff
@@ -37,7 +37,6 @@
         description:
         - The name of the VM to be used to create template.
         type: str
-        required: true
```
anna-savina, line 40: I think it's worth mentioning that `vm_name` is required when the state is `present`. Maybe not here, but in the description.

mikemorency (approved):

anna-savina:
Remove `vm_name` from `vmware_content_template` integration test to test changes, please

https://github.com/ansible-collections/vmware.vmware/blob/8ad8eb486af67dc5b439f2ae3bea992adb35efa0/tests/integration/targets/vmware_content_template/tasks/main.yml#L72C1-L72C28

anna-savina (approved):

PR #104, opened 2025-01-22
migrate community vmware_folder module to here

mikemorency:
##### SUMMARY
This is migrating the community.vmware.vmware_folder module to here. The functionality is the same but the inputs have changed slightly to allow for more flexible usage.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
folder

### DIFF
--- a/plugins/modules/folder.py
+++ b/plugins/modules/folder.py
@@ -11,7 +11,7 @@ __metaclass__ = type
 
 DOCUMENTATION = r'''
 ---
-module: cluster
+module: folder
 short_description: Manage VMware vSphere folders
 description:
     - Adds or removes VMware vSphere folders. This module does not manage folders inside of datastores.
@@ -24,7 +24,7 @@ options:
             - The name of the datacenter where the folder should be created.
             - Only used if the O(relative_path) option is used.
         type: str
-        required: true
+        required: false
         aliases: [datacenter_name]
     folder_type:
         description:
@@ -33,6 +33,7 @@ options:
             - For example, a folder at path /DC-01/vm/my/folder has folder type 'vm'.
             - Only used if the O(relative_path) option is used.
         type: str
+        required: false
         choices: [vm, host, storage, network]
     relative_path:
         description:
@@ -275,8 +276,8 @@ def main():
 
         new_folder = vmware_folder.create_folder()
         module.exit_json(changed=True, folder={
-                'moid': vmware_folder.folder_object._GetMoId(),
-                'name': vmware_folder.folder_object.name
+                'moid': new_folder._GetMoId(),
+                'name': new_folder.name
             })
 
     if module.params['state'] == 'absent':
--- a/plugins/modules/folder.py
+++ b/plugins/modules/folder.py
@@ -199,7 +199,7 @@ class VmwareFolder(ModulePyvmomiBase):
             The final folder object created
         """
         split_path = self.absolute_folder_path.split('/')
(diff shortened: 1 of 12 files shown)


plugins/modules/folder.py
```diff
@@ -0,0 +1,292 @@
...
+
+- name: Delete The Folder and All Of Its Contents
+  vmware.vmware.folder:
+    hostname: "https://vcenter"
+    username: "username"
+    password: "password"
+    validate_certs: false
+    absolute_path: /DC01/storage/my
+    state: absent
+
```
anna-savina, line 101: By default `remove_vm_data` is false
```suggestion
remove_vm_data: true
```

plugins/modules/folder.py
```diff
@@ -0,0 +1,292 @@
...
+        choices: [vm, host, storage, network]
+    relative_path:
+        description:
+            - The relative path of the folder to create. The relative path should not include the datacenter or the folder type.
+            - For example the relative path for the folder /DC-01/vm/my/folder is my/folder
+            - One of O(relative_path) or O(absolute_path) must be specified.
+        type: str
+    absolute_path:
+        description:
+            - The absolute path of the folder to create. The absolute path should include the datacenter or the folder type.
```
anna-savina, line 46:
```suggestion
            - The absolute path of the folder to create. The absolute path should include the datacenter and the folder type.
```

plugins/modules/folder.py
```diff
@@ -0,0 +1,292 @@
...
+        type: str
+    state:
+        description:
+            - Create V(present) or remove V(absent) a vSphere folder.
+            - When state is V(present), all folders in the specified path will be created if they are missing.
+            - When state is V(absent), all content inside the folder will be deleted. See O(remove_vm_data)
+        choices: [ absent, present ]
+        default: present
+        type: str
+    remove_vm_data:
```
anna-savina, line 58: I think this argument needs more testing:
- folder can't be deleted if VM is powered on, but module doesn't report failure
- I'm not sure that `remove_vm_data: false` works, as I can see the VM is always deleted.

tests/integration/targets/vmware_folder/vars/main.yml
```diff
@@ -0,0 +1,3 @@
+---
```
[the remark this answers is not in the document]
mikemorency: vars/ are not meant to be overridden by the user, while defaults can be. in this case the two variables here are used a lot in the role and i want to define them in one place, but they shouldnt really change for simulator/eco

plugins/modules/folder.py
mikemorency: I tested and fixed the issue when a VM is powered on. The module will throw a failure now

I manually tested and verified `remove_vm_data` is working as expected. I cant add it to the automated tests though. Theres no module to query the datastore file contents and no module to delete a directory on a datastore

anna-savina (approved):

mariolenz:
> This is migrating the community.vmware.vmware_folder module to here. The functionality is the same but the inputs have changed slightly to allow for more flexible usage.

@mikemorency excuse me, but I think there's no `community.vmware.vmware_folder`. Did you mean `community.vmware.vcenter_folder`?

mikemorency:
Ah, yes I did mean vcenter_folder

mariolenz:
Thanks for the clarification! Now I know what module to deprecate :-)

Although I think I'll work on ansible-collections/community.vmware#2337 / #128 first.

PR #107, opened 2025-01-24
add event counting file

mikemorency:
##### SUMMARY
This adds an event_counting.yml file to the repo. This file is still experimental, but counts and identifies the number of nodes indirectly managed by a module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
event_counting


meta/event_counting.yml
```diff
@@ -0,0 +1,31 @@
...
+        ansible_vmware_instance_uuid: .instance_uuid,
+        ansible_vmware_bios_uuid: .hw_product_uuid,
+        ansible_vmware_moid: .moid
+      },
+      facts: {
+        device_type: "VM"
+      }
+    }
+
+vm_list_group_by_clusters_info:
```
AlanCoding, line 17:
```suggestion
vmware.vmware.vm_list_group_by_clusters_info:
```

I think this is obviously more consistent with the other entry.

mikemorency: yes, good catch :)

pb82:
@mikemorency Ansible Core asked us to move the file into `extensions` instead of `meta`.

AlanCoding:
The extensions folder needs to have things in subfolders. We were talking about `extensions/audit/event_query.yml`.

extensions/audit/event_query.yml
```diff
@@ -0,0 +1,31 @@
+---
+# VMs
+vmware.vmware.guest_info:
```
pb82, line 3: This may be me not understanding the collection, but why are we running the query for `guest_info` instead of `vmware.vmware.vmware_guest`? Isn't the latter used to create VMs?

mikemorency: `vmware.vmware.vmware_guest` doesnt exist in this collection. The `vmware_guest` module is still in the community collection, `community.vmware`

mikemorency: we have plans to migrate `vmware_guest` to this repo eventually, and we can add it to the query file then

pb82: is `guest_info` then used in the same way and allows you to create VMs?

pb82 (approved):
Verified with the controller indirect host counting implementation.

anna-savina (approved):

PR #109, opened 2025-01-31
Migrate/content library

mikemorency:
##### SUMMARY
Migrates the community vmware_content_library_manager functionality here.
I decided to split this module into two, one for local libraries and one for subscribed libraries. The module was getting too complex trying to fit them both in one.
I also added functionality for authentication and publishing local libraries which was not included in the original.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
local_content_library
subscribed_content_library

##### ADDITIONAL INFORMATION
The subscribed library module can only be tested if theres a library to subscribe to, so i combined their integration tests into one. I figured that was OK since both are still being tested thoroughly


plugins/modules/local_content_library.py
```diff
@@ -0,0 +1,406 @@
...
+        description:
+            - The description for the content library.
+        type: str
+        aliases: [library_description]
+        default: ''
+    datastore:
+        description:
+            - The name of the datastore that should be a storage backing for the library.
+            - This parameter is required when O(state) is V(present)
+            - This parameter only takes affect when the library is first created. You cannot change the
```
anna-savina, line 45:
```suggestion
            - This parameter only takes effect when the library is first created. You cannot change the
```

plugins/modules/local_content_library.py
```diff
@@ -0,0 +1,406 @@
...
+            create_spec
+        )
+
+    def format_library_for_output(self, library_id=None, library=None):
+        """
+            This is a helper function to enrich the module output from library attributes.
+            We need to add the publish URL to the output epending on if the library has been published or not.
+        """
+        if not library_id and not library:
+            self.module.fail_json('You need to supply either the library object or ID to format them for output')
```
anna-savina, line 334:
```suggestion
            self.module.fail_json(msg='You need to supply either the library object or ID to format them for output')
```

plugins/modules/subscribed_content_library.py
```diff
@@ -0,0 +1,318 @@
...
+            - The description for the content library.
+        type: str
+        required: false
+        aliases: [library_description]
+        default: ''
+    datastore:
+        description:
+            - The name of the datastore that should be a storage backing for the library.
+            - This parameter is required when O(state) is V(present)
+            - This parameter only takes affect when the library is first created. You cannot change the
```
anna-savina, line 46:
```suggestion
            - This parameter only takes effect when the library is first created. You cannot change the
```

plugins/modules/subscribed_content_library.py
```diff
@@ -0,0 +1,318 @@
...
+        description:
+            - Whether the content library should be present or absent.
+        type: str
+        default: present
+        choices: [present, absent]
+
+    authentication_method:
+        description:
+            - The method of authentication to use if this is a subscribed or published local library.
+            - If the library is a local, non-published library, this option has no affect.
```
anna-savina, line 61:
```suggestion
            - If the library is a local, non-published library, this option has no effect.
```

anna-savina (approved):

PR #112, opened 2025-02-05
rename folder_template_from_vm module

mikemorency:
##### SUMMARY
Renaming `folder_template_from_vm` to `folder_template` for two reasons:
1. the new name is more consistent with a similar module, `content_template`
2. this module manages the template's existence, and I think the new name better reflects that

I have another module coming that will deploy a vm from a folder template, which will be called `deploy_folder_template`

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
folder_template

##### ADDITIONAL INFORMATION
I think a major release is going to happen soon-ish so I set this module redirect to expire with version 3

### DIFF
--- a/changelogs/fragments/112-rename-folder-template-module.yml
+++ b/changelogs/fragments/112-rename-folder-template-module.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - folder_template - renamed folder_template_from_vm to folder_template, in order to match content_template module and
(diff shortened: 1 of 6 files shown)


mikemorency:
@mariolenz  can you please review and let me know what you think? 1. im hesitant about any breaking changes and 2. I want to make sure I set the redirect up correctly

PR #113, opened 2025-02-06
add module to deploy vm from folder template

mikemorency:
##### SUMMARY
Adding a module to deploy a vm from  a folder template (instead of a content library template).
This is the third module to deploy something, and they were all introduced in this release. So this PR includes some small changes to the other two modules to make them all consistent

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
deploy_folder_template

##### ADDITIONAL INFORMATION
This is a net new module, and replaces some of the functionality from vmware_guest


plugins/modules/deploy_folder_template.py
```diff
@@ -0,0 +1,266 @@
...
+        type: str
+        required: False
+    esxi_host:
+        description:
+            - The name of the ESXi host onto which the virtual machine should be deployed.
+            - If O(esxi_host) and O(resource_pool) are both specified, O(resource_pool) must belong to O(esxi_host).
+            - If O(esxi_host) and O(cluster) are both specified, O(esxi_host) must be a member of O(cluster).
+            - Changing this option will not result in the VM being redeployed (it does not affect idempotency).
+        type: str
+        aliases: [host]
```
anna-savina, line 55:
```suggestion
        required: False
        aliases: [host]
```

plugins/modules/deploy_content_library_ovf.py
```diff
@@ -194,32 +144,20 @@
...
-                self.params['resource_pool'],
-                fail_on_missing=True
-            )._GetMoId()
-        cluster = self.get_cluster_by_name_or_moid(
-            self.params['cluster'],
-            fail_on_missing=True,
-            datacenter=self.datacenter
-        )
-        return cluster.resource_pool._GetMoId()
+        self._library_item_id = self.params.get('library_item_id')
```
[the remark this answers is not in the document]
mikemorency: this is intentional
`library_item_id` is not supposed to set by a user/dev, it is supposed to be a calculated value. To achieve this, the variable uses special getter and setter methods, as described in this article: https://www.geeksforgeeks.org/getter-and-setter-in-python/

plugins/modules/deploy_content_library_ovf.py
```diff
@@ -194,32 +144,20 @@
...
 
     def create_deploy_spec(self):
         deployment_target = OvfLibraryItems.DeploymentTarget(
-            resource_pool_id=self.resource_pool_id,
-            folder_id=self.get_deployment_folder()._GetMoId()
+            resource_pool_id=(self.resource_pool._GetMoId() if self.datastore else None),
```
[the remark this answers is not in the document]
mikemorency: oh, good catch! the resource pool should always be defined here, so no if/else

plugins/modules/deploy_content_library_template.py
```diff
@@ -205,40 +155,33 @@
...
 
-class VmwareContentDeployTemplate(VmwareContentDeploy):
+class VmwareContentDeployTemplate(ModuleVmDeployBase):
     def __init__(self, module):
         """Constructor."""
         super(VmwareContentDeployTemplate, self).__init__(module)
 
+        self.rest_base = ModuleRestBase(module)
         self.template_service = self.rest_base.api_client.vcenter.vm_template.LibraryItems
+        self._library_item_id = self.params.get('library_item_id')
```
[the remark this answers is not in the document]
mikemorency: same explanation as with the OVF module

plugins/modules/deploy_content_library_ovf.py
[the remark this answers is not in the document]
anna-savina: Related question: should it be used `self._library_item_id` in line - 
https://github.com/ansible-collections/vmware.vmware/blob/090c011b10c6b28450728f97024418646d5b2e11/plugins/modules/deploy_content_library_ovf.py#L177

if yes - change all occurrences of the `self._library_item_id`  in a code, please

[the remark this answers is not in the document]
anna-savina: ok. thank you for the explanation

plugins/modules/deploy_content_library_template.py
```diff
@@ -15,7 +15,7 @@
 module: deploy_content_library_template
 short_description: Deploy a virtual machine from a template in a content library.
 description:
-    - Create or destroy a VM based on a template in a content library.
+    - Create a VM based on a template in a content library.
```
anna-savina, line 18: Make the same changes for `deploy_content_library_ovf ` the module please

plugins/modules/deploy_folder_template.py
```diff
@@ -0,0 +1,268 @@
...
+    hostname: "https://vcenter"
+    username: "username"
+    password: "password"
+    datacenter: "my-datacenter"
+    vm_name: "my_vm"
+    vm_name_match: "first"
+    template_name: "my_template"
+    template_folder: "nested/folder/path/templates"
+
+- name: Destroy A Template In A Folder
```
anna-savina, line 99: No absent state

plugins/modules/deploy_folder_template.py
```diff
@@ -0,0 +1,268 @@
...
+
+options:
+    template_name:
+        description:
+            - The name of the template to use when deploying.
+            - You must also supply O(template_folder) if you use this parameter.
+            - This parameter is not used if O(template_id) is supplied.
+        type: str
+        required: False
+    template_folder:
```
anna-savina, line 35: Maybe it's worth providing two options: template_folder_name and template_folder_id?
Not necessarily, just thoughts

plugins/modules/deploy_folder_template.py
```diff
@@ -0,0 +1,268 @@
...
+            self.module.fail_json(msg="Object found matching template name or ID is not a template")
+
+        self.template = template
+        return template
+
+    def deploy(self, deploy_spec):
+        try:
+            task = self.template.Clone(name=self.params['vm_name'], folder=self.vm_folder, spec=deploy_spec)
+            _, task_result = RunningTaskMonitor(task).wait_for_completion()   # pylint: disable=disallowed-name
+        except TaskError as e:
```
anna-savina, line 195: Maybe add also `except Exception as err` in case another exceptions occur.

tests/integration/targets/vmware_deploy_folder_template/tasks/eco-vcenter.yml
```diff
@@ -0,0 +1,81 @@
...
+    - name: Deploy VM
+      vmware.vmware.deploy_folder_template:
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        validate_certs: false
+        datacenter: "{{ vcenter_datacenter }}"
+        datastore: "{{ shared_storage_02 }}"
+        template_name: "{{ test_template_name }}"
+        template_folder: "{{ test_template_folder }}"
```
anna-savina, line 50: This folder doesn't exist at vCenter and neither does the template. Should it be like that? I'm not sure if the test covered module functionality.

plugins/modules/deploy_folder_template.py
mikemorency: yea thats a good idea, ill add it

tests/integration/targets/vmware_deploy_folder_template/tasks/eco-vcenter.yml
anna-savina: but now `template_folder` is not in use at all. How is it different from the `deploy_content_library_template` module since the template is located in `ansible_ci_test_resources` content library?

mikemorency: > since the template is located in ansible_ci_test_resources content library
The content library just stores the template in the datacenter root folder, like a normal template. You can access it either way for whatever reason

I created a new folder and copied the template there. Then I updated the test cases to use `template_folder` and then `template_id`, so both are checked now :+1:

anna-savina (approved):

PR #114, opened 2025-02-10
Migrate/vmware guest powerstate

OrrBG:
##### SUMMARY
Migrate the vmware_guest_powerstate module

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
vmware_guest_powerstate

### DIFF
--- a/plugins/modules/vm_powerstate.py
+++ b/plugins/modules/vm_powerstate.py
@@ -1,0 +1,470 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2025, Ansible Cloud Team (@ansible-collections)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: vm_powerstate
+short_description: Manages power states of virtual machines in vCenter
+description:
+    - Manages power states of virtual machines in vCenter, e.g., Power on / Power off / Restart.
+author:
+    - Ansible Cloud Team (@ansible-collections)
+
+options:
+    datacenter:
+        description:
+            - The datacenter where the VM you'd like to operate the power.
+        type: str
+        required: true
+    state:
+        description:
+            - Set the state of the virtual machine.
+        choices: [ powered-off, powered-on, reboot-guest, restarted, shutdown-guest, suspended]
+        default: powered-on
+        type: str
+    name:
+        description:
+            - Name of the virtual machine to work with.
+            - Virtual machine names in vCenter are not necessarily unique, which may be problematic, see O(name_match).
+            - This is required if O(moid) or O(uuid) is not supplied.
+        type: str
+    name_match:
+        description:
+            - If multiple virtual machines matching the name, use the first or last found.
+        default: first
+        choices: [ first, last ]
+        type: str
+    uuid:
+        description:
+            - UUID of the instance to manage if known, this is VMware's unique identifier.
+            - This is required if O(name) or O(moid) is not supplied.
+        type: str
+    moid:
+        description:
+            - Managed Object ID of the instance to manage if known, this is a unique identifier only within a single vCenter instance.
+            - This is required if O(name) or O(uuid) is not supplied.
+        type: str
+    use_instance_uuid:
+        description:
+            - Whether to use the VMware instance UUID rather than the BIOS UUID.
+        default: false
+        type: bool
+    folder:
+        description:
+            - Destination folder, absolute or relative path to find an existing guest.
+            - Should be the full folder path, with or without the 'datacenter/vm/' prefix
+            - For example 'datacenter_name/vm/path/to/folder' or 'path/to/folder'
+        type: str
+        required: false
+    scheduled_at:
+        description:
+            - Date and time in string format at which specified task needs to be performed.
+            - "The required format for date and time - 'dd/mm/yyyy hh:mm'."
+            - Scheduling task requires vCenter server. A standalone ESXi server does not support this option.
+        type: str
+        required: false
+    scheduled_task_name:
+        description:
+            - Name of scheduled task.
+            - Valid only if O(scheduled_at) is specified.
+        type: str
+        required: false
+    scheduled_task_description:
+        description:
+            - Description of scheduled task.
+            - Valid only if O(scheduled_at) is specified.
+        type: str
+        required: false
+    scheduled_task_enabled:
+        description:
+            - Flag to indicate whether the scheduled task is enabled or disabled.
+        type: bool
+        default: true
+    force:
+        description:
+            - Ignore warnings and complete the actions.
+            - This parameter is useful while forcing virtual machine state.
+        default: false
+        type: bool
+    timeout:
+        description:
+            - If this argument is set to a positive integer, the module will wait for the VM to reach the poweredoff state.
+            - The value sets a timeout in seconds for the module to wait for the state change.
+            - Can not be used if O(state) is V(shutdown-guest) or V(reboot-guest).
+        default: 3600
+        type: int
+    question_answers:
+        description:
+            - A list of questions to answer, should one or more arise while waiting for the task to complete.
+            - Some common uses are to allow a cdrom to be changed even if locked, or to answer the question as to whether a VM was copied or moved.
+            - Can be used if O(state) is V(powered-on).
+        suboptions:
+            question:
+                description:
+                    - The message id, for example C(msg.uuid.altered).
+                type: str
+                required: true
+            response:
+                description:
+                    - The choice key, for example C(button.uuid.copiedTheVM).
+                type: str
+                required: true
+        type: list
+        elements: dict
+        required: false
+
+
+extends_documentation_fragment:
+    - vmware.vmware.base_options
+'''
+
+EXAMPLES = r'''
+- name: Set the state of a virtual machine to poweroff
+  vmware.vmware.vm_powerstate:
+    hostname: "{{ vcenter_hostname }}"
+    username: "{{ vcenter_username }}"
+    password: "{{ vcenter_password }}"
+    datacenter: "{{ vm_datacenter }}"
+    folder: "/{{ datacenter_name }}/vm/my_folder"
+    name: "{{ guest_name }}"
+    state: powered-off
+  register: deploy
+
+- name: Set the state of a virtual machine to poweron using MoID
+  vmware.vmware.vm_powerstate:
+    hostname: "{{ vcenter_hostname }}"
+    username: "{{ vcenter_username }}"
+    password: "{{ vcenter_password }}"
+    datacenter: "{{ vm_datacenter }}"
+    folder: "/{{ datacenter_name }}/vm/my_folder"
+    moid: vm-42
+    state: powered-on
+  register: deploy
+
+- name: Set the state of a virtual machine to poweroff at given scheduled time
+  vmware.vmware.vm_powerstate:
+    hostname: "{{ vcenter_hostname }}"
+    username: "{{ vcenter_username }}"
+    password: "{{ vcenter_password }}"
+    datacenter: "{{ vm_datacenter }}"
+    folder: "/{{ datacenter_name }}/vm/my_folder"
+    name: "{{ guest_name }}"
+    state: powered-off
+    scheduled_at: "09/01/2018 10:18"
+    scheduled_task_name: "task_00001"
+    scheduled_task_description: "Sample task to poweroff VM"
+    scheduled_task_enabled: true
+  register: deploy_at_scheduled_datetime
+
+- name: Wait for the virtual machine to shutdown
+  vmware.vmware.vm_powerstate:
+    hostname: "{{ vcenter_hostname }}"
+    username: "{{ vcenter_username }}"
+    password: "{{ vcenter_password }}"
+    datacenter: "{{ vm_datacenter }}"
+    name: "{{ guest_name }}"
+    state: shutdown-guest
+  register: deploy
+
+- name: Automatically answer if a question locked a virtual machine
+  block:
+    - name: Power on a virtual machine without the answer param
+      vmware.vmware.vm_powerstate:
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        datacenter: "{{ vm_datacenter }}"
+        validate_certs: false
+        folder: "{{ f1 }}"
+        name: "{{ vm_name }}"
+        state: powered-on
+  rescue:
+    - name: Power on a virtual machine with the answer param
+      vmware.vmware.vm_powerstate:
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        datacenter: "{{ vm_datacenter }}"
+        validate_certs: false
+        folder: "{{ f1 }}"
+        name: "{{ vm_name }}"
+        question_answers:
+          - question: "msg.uuid.altered"
+            response: "button.uuid.copiedTheVM"
+        state: powered-on
+'''
+
+RETURN = r'''
+vm:
+    description:
+        - Information about the target VM
+    returned: On success
+    type: dict
+    sample:
+        moid: vm-79828,
+        name: test-d9c1-vm
+'''
+
+try:
+    from pyVmomi import vim, vmodl
+except ImportError:
+    pass
+
+from random import randint
+from datetime import datetime
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils._text import to_text, to_native
+
+from ansible_collections.vmware.vmware.plugins.module_utils._module_pyvmomi_base import (
+    ModulePyvmomiBase
+)
+from ansible_collections.vmware.vmware.plugins.module_utils._vmware_argument_spec import (
+    base_argument_spec
+)
+from ansible_collections.vmware.vmware.plugins.module_utils._vmware_folder_paths import format_folder_path_as_vm_fq_path
+from ansible_collections.vmware.vmware.plugins.module_utils._vmware_tasks import TaskError, RunningTaskMonitor, VmQuestionHandler
+
+
+class VmPowerstateModule(ModulePyvmomiBase):
+    def __init__(self, module):
+        super(VmPowerstateModule, self).__init__(module)
+
+        self.result = dict(
+            changed=False,
+            vm=dict(
+                name=None,
+                moid=None
+            )
+        )
+
+        vm_list = self.get_vm_using_params(fail_on_missing=True)
+        self.vm = vm_list[0]
+        state = self.module.params['state']
+        self.desired_state = state.replace('_', '').replace('-', '').lower()
+        self.current_state = self.vm.summary.runtime.powerState.lower()
+        self.result["vm"]['moid'] = self.vm._GetMoId()
+        self.result["vm"]['name'] = self.vm.name
+
+    def standardize_folder_param(self):
+        """
+        Removes all "/" characters in the folder path if the folder param exists
+        """
+        if self.module.params['folder']:
+            self.module.params['folder'] = format_folder_path_as_vm_fq_path(
+                self.module.params['folder'],
+                self.module.params['datacenter']
+            )
+
+    def run_vm_configuration_task(self, task):
+        if not task:
+            return
+        try:
+            succeeded, task_result = RunningTaskMonitor(task).wait_for_completion(
+                vm=self.vm, timeout=self.module.params['timeout'], answers=self.module.params.get('question_answers', None))
+        except TaskError as e:
+            self.module.fail_json(msg=to_text(e))
+        finally:
+            self.result['changed'] = True
+
+    def reset_vm(self):
+        if self.current_state not in ('poweredon', 'poweringon', 'resetting', 'poweredoff'):
+            self.module.fail_json(msg="Cannot restart virtual machine in the current state %s" % self.current_state)
+        return self.vm.Reset()
+
+    def suspend_vm(self):
+        if self.current_state not in ('poweredon', 'poweringon'):
+            self.module.fail_json(msg='Cannot suspend virtual machine in the current state %s' % self.current_state)
+        return self.vm.Suspend()
+
+    def set_vm_state_using_vmtools(self):
+        task = None
+        if self.current_state == 'poweredon':
+            if self.vm.guest.toolsRunningStatus == 'guestToolsRunning':
+                if self.desired_state == 'shutdownguest':
+                    task = self.vm.ShutdownGuest()
+                else:
+                    task = self.vm.RebootGuest()
+                # Set result['changed'] immediately because
+                # shutdown and reboot return None.
+                self.result['changed'] = True
+            else:
+                self.module.fail_json(msg="VMware tools should be installed for guest shutdown/reboot")
+        elif self.current_state == 'poweredoff' and self.desired_state == 'shutdownguest':
+            self.result['changed'] = False
+        else:
+            self.module.fail_json(msg="Virtual machine %s must be in poweredon state for guest reboot" % self.vm.name)
+        return task
+
+    def set_vm_powerstate(self):
+        """
+        Set the power status for a VM determined by the current and
+        requested states. force is forceful
+        """
+        # Need Force
+        if not self.module.params['force'] and self.current_state not in ['poweredon', 'poweredoff']:
+            self.module.fail_json(msg="Virtual Machine is in %s power state. Force is required!" % self.current_state)
+
+        task = None
+        desired_powerstate = {
+            'poweredoff': self.vm.PowerOff,
+            'poweredon': self.vm.PowerOn,
+            'rebootguest': self.set_vm_state_using_vmtools,
+            'restarted': self.reset_vm,
+            'shutdownguest': self.set_vm_state_using_vmtools,
+            'suspended': self.suspend_vm,
+        }
+        try:
+            if self.desired_state in desired_powerstate:
+                task = desired_powerstate[self.desired_state]()
+            else:
+                self.module.fail_json(msg="Unsupported expected state provided: %s" % self.desired_state)
+
+        except Exception as e:
+            self.module.fail_json(msg=to_text(e))
+
+        self.run_vm_configuration_task(task)
+
+    def configure_vm_powerstate(self):
+        """
+        Configures a VMs powerstate
+        """
+        scheduled_at = self.module.params.get('scheduled_at', None)
+        if scheduled_at:
+            scheduled_task_spec = self.configure_scheduled_task_spec(scheduled_at)
+            self.configure_vm_scheduled_powerstate(scheduled_task_spec)
+        else:
+            self.set_vm_powerstate()
+
+    def configure_scheduled_task_spec(self, scheduled_at):
+        """
+        Returns:
+            ScheduledTaskSpec, object that contains all specifications regarding the scheduled task
+        """
+        if not self.is_vcenter():
+            self.module.fail_json(msg="Scheduling task requires vCenter, hostname %s "
+                                  "is an ESXi server." % self.module.params.get('hostname'))
+        powerstate = {
+            'powered-off': vim.VirtualMachine.PowerOff,
+            'powered-on': vim.VirtualMachine.PowerOn,
+            'reboot-guest': vim.VirtualMachine.RebootGuest,
+            'restarted': vim.VirtualMachine.Reset,
+            'shutdown-guest': vim.VirtualMachine.ShutdownGuest,
+            'suspended': vim.VirtualMachine.Suspend,
+        }
+        try:
+            scheduled_date = datetime.strptime(scheduled_at, '%d/%m/%Y %H:%M')
+        except ValueError as e:
+            self.module.fail_json(msg="Failed to convert given date and time string to Python datetime object,"
+                                  "please specify string in 'dd/mm/yyyy hh:mm' format: %s" % to_native(e))
+        scheduled_task_spec = vim.scheduler.ScheduledTaskSpec()
+        scheduled_task_spec.name = self.module.params['scheduled_task_name'] or 'task_%s' % str(randint(10000, 99999))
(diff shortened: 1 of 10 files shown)


plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+            - 'Examples:'
+            - '   folder: /ha-datacenter/vm'
+            - '   folder: ha-datacenter/vm'
+            - '   folder: /datacenter1/vm'
+            - '   folder: datacenter1/vm'
+            - '   folder: /datacenter1/vm/folder1'
+            - '   folder: datacenter1/vm/folder1'
+            - '   folder: /folder1/datacenter1/vm'
+            - '   folder: folder1/datacenter1/vm'
+            - '   folder: /folder1/datacenter1/vm/folder2'
```
mikemorency, line 73: which module did you copy this from? we should update that module too.

Remove all of the examples, remove `The folder should include the datacenter. ESX's datacenter is ha-datacenter`, add two lines like
```
- Should be the full folder path, with or without the 'datacenter/vm/' prefix
- For example 'datacenter name/vm/path/to/folder' or 'path/to/folder'
```

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+        type: str
+        required: false
+    scheduled_at:
+        description:
+            - Date and time in string format at which specified task needs to be performed.
+            - "The required format for date and time - 'dd/mm/yyyy hh:mm'."
+            - Scheduling task requires vCenter server. A standalone ESXi server does not support this option.
+        type: str
+        required: false
+    schedule_task_name:
```
mikemorency, line 83: i think `scheduled_task_name` sounds more natural

mikemorency: all of the scheduled task options should be updated similarly

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+        description:
+            - Name of schedule task.
```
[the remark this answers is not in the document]
mikemorency: all of the scheduled task descriptions should be updated similarly

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+        default: false
+        type: bool
+    state_change_timeout:
+        description:
+            - If the O(state=shutdown-guest), by default the module will return immediately after sending the shutdown signal.
+            - If this argument is set to a positive integer, the module will instead wait for the VM to reach the poweredoff state.
+            - The value sets a timeout in seconds for the module to wait for the state change.
+        default: 0
+        type: int
+    answer:
```
mikemorency, line 113: i think `question_answers` is slightly less vague

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+        type: list
+        elements: dict
+        required: false
+
+
+extends_documentation_fragment:
+    - vmware.vmware.base_options
+'''
+
+EXAMPLES = r'''
```
mikemorency, line 138: need to update examples, they all say community

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+        validate_certs: false
+        folder: "{{ f1 }}"
+        name: "{{ vm_name }}"
+        answer:
+          - question: "msg.uuid.altered"
+            response: "button.uuid.copiedTheVM"
+        state: powered-on
+'''
+
+RETURN = r'''
```
mikemorency, line 209: need to update the results. Ill comment further down but you should only return the vm name, moid, and maybe the new state

mikemorency: oh, also the "task" details. like here https://github.com/ansible-collections/vmware.vmware/blob/main/plugins/modules/esxi_maintenance_mode.py#L228
`task_result['result']` will be the VM object you modified, which cannot be passed directly to ansible output. You can just `del task_result['result']` before you output it though

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+
+try:
+    from pyVmomi import vim, vmodl
+except ImportError:
+    pass
+
+from random import randint
+from datetime import datetime
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.vmware.plugins.module_utils import vmware
```
mikemorency, line 310: you cant import from community

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+from ansible_collections.vmware.vmware.plugins.module_utils._module_pyvmomi_base import (
+    ModulePyvmomiBase
+)
+
+
+class VmwareGuestPowerstateModule(ModulePyvmomiBase):
+    def __init__(self, module):
+        super(VmwareGuestPowerstateModule, self).__init__(module)
+
+        self.result = dict(changed=False)
```
mikemorency, line 322:
```suggestion
        self.result = dict(changed=False, vm=dict(name=self.params['name'], moid=None))
```

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+    def get_vm(self):
+        """
+        Finds and gets VM according to the specified name/uuid/moid
+        Returns:
+            tuple, vm object and PyVmomi object if found 
+        """
+        if self.module.params['folder']:
+            self.module.params['folder'] = self.module.params['folder'].rstrip('/')
+
+        pyv = vmware.PyVmomi(self.module)
```
mikemorency, line 333: pyvmomi is already initialized as part of your base class

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+)
+
+
+class VmwareGuestPowerstateModule(ModulePyvmomiBase):
+    def __init__(self, module):
+        super(VmwareGuestPowerstateModule, self).__init__(module)
+
+        self.result = dict(changed=False)
+
+    def get_vm(self):
```
mikemorency, line 324: this should be broken into two:
1. format the folder that comes from the user. The user can provide an absolute or relative folder path. Refer to this for an example, https://github.com/ansible-collections/vmware.vmware/blob/main/plugins/modules/folder_template_from_vm.py#L172-L176
2. Lookup the VM. There is a method in the base class that can be used to lookup a vm from whatever parameters the user supplied, https://github.com/ansible-collections/vmware.vmware/blob/main/plugins/module_utils/_module_pyvmomi_base.py#L109

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+            else:
+                self.result = vmware.set_vm_power_state(pyv.content, vm, self.module.params['state'], self.module.params['force'], self.module.params['state_change_timeout'],
+                                            self.module.params['answer'])
+            self.result['answer'] = self.module.params['answer']
+        
+        self.result['moid'] = vm._GetMoId()
+        self.result['name'] = vm.name
+        self.result['instance'] = vmware.gather_vm_facts(pyv.content, vm)
+
+    def configure_vm_scheduled_powerstate(self, vm, pyv, scheduled_at):
```
mikemorency, line 363: too complex, please break it up. Id probably make a method to create the task spec, a method to apply the task spec/start the task, and probably move the `powerstate` map to a class or global variable since its not dependent on the class instance and never changes

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+
+        self.result['changed'] = True
+    
+    def current_state_matches_desired_state(self, vm, pyv):
+        """
+        Checks the hosts current power state and compares it to the desired power state setting.
+        Returns:
+            bool, true if they match, otherwise false
+        """
+        facts = vmware.gather_vm_facts(pyv.content, vm)
```
mikemorency, line 434: so when you lookup the vm in one of the methods above, it returns an object that has all of the properties you need. You dont need to "gather facts" again.
For example, 
```
vm = self.get_vm_using_params(whatever)
if vm.summary.runtime.powerState == 'poweredon':
```

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+
+        vm = pyv.get_vm()
+
+        if not vm:
+            id = self.module.params.get('uuid') or self.module.params.get('moid') or self.module.params.get('name')
+            self.module.fail_json(msg="Unable to set power state for non-existing virtual machine : '%s'" % id)
+
+        return (vm, pyv)
+
+    def configure_vm_powerstate(self, vm, pyv):
```
mikemorency, line 343: this method will need to change pretty significantly since you cant import from community. I'd suggest focusing on getting `set_vm_power_state` squared away first, and then coming back to the answerable questions once the module is nearing completion.
Community has the `set_vm_power_state` in the module_utils but I dont think that is necessary, you can just create a similar method in this file. I think theres a lot of opportunity to simplify that method, but I wanted to draw attention to the "task" that is created when a VM changes state. The task needs to be monitored for completion. We have methods for that, as seen here:
https://github.com/ansible-collections/vmware.vmware/blob/main/plugins/modules/esxi_maintenance_mode.py#L158C1-L175C27

plugins/modules/vmware_guest_powerstate.py
mikemorency: heres an example of task output
https://github.com/ansible-collections/vmware.vmware/blob/main/plugins/modules/esxi_maintenance_mode.py#L90

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,510 @@
...
+        result['changed'] = True
+        module.exit_json(**result)
+    
+    vmware_guest_powerstate.configure_vm_powerstate(vm, pyv)
+    result = vmware_guest_powerstate.result
+    
+    if result.get('failed') is True:
+        module.fail_json(**result)
+
+    module.exit_json(**result)
```
mikemorency, line 506: i think the overall flow of this is good. You can add the vm name/moid to output here if that is easier.
You could break `configure_vm_powerstate` up and add the different options here, if that makes things simpler. for example
```
if scheduling:
    vmware_guest_powerstate.configure_powerstate_schedule()
elif state == poweron
   vmware_guest_powerstate.power_on()
elif state ==poweroff
  vmware_guest_powerstate.power_off()
elif state ==restart
  vmware_guest_powerstate.restart()

File: plugins/modules/vmware_guest_powerstate.py (hunk 1 above)
Comment (OrrBG) [reply plugins/modules/vmware_guest_powerstate.py:73]: I copied this from the corresponding community module

File: tests/integration/targets/vmware_guest_powerstate/tasks/main.yml [hunk 15]
```diff
@@ -0,0 +1,231 @@
...
+        password: "{{ vcenter_password }}"
+        validate_certs: false
+        folder: "{{ vm_folder }}"
+        name: "{{ vm }}"
+        question_answers:
+          - question: "msg.uuid.altered"
+            response: "button.uuid.copiedTheVM"
+        state: powered-on
+
+  always:
```
mikemorency, line 218: you need a task to delete your resource pool,
```
    - name: Delete test resource pool
      vmware.vmware_rest.vcenter_resourcepool:
        vcenter_hostname: "{{ vcenter_hostname }}"
        vcenter_username: "{{ vcenter_username }}"
        vcenter_password: "{{ vcenter_password }}"
        vcenter_validate_certs: false
        name: "{{ test_resource_pool }}"
        parent: "{{ vcenter_rp_id }}"
        state: absent
```

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+        default: true
+    force:
+        description:
+            - Ignore warnings and complete the actions.
+            - This parameter is useful while forcing virtual machine state.
+        default: false
+        type: bool
+    timeout:
+        description:
+            - If the O(state=shutdown-guest), by default the module will return immediately after sending the shutdown signal.
```
mikemorency, line 99:
```suggestion
            - If the O(state) is V(shutdown-guest), by default the module will return immediately after sending the shutdown signal.
```

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+            - If this argument is set to a positive integer, the module will instead wait for the VM to reach the poweredoff state.
+            - The value sets a timeout in seconds for the module to wait for the state change.
+        default: 0
+        type: int
+    question_answers:
+        description:
+            - A list of questions to answer, should one or more arise while waiting for the task to complete.
+            - Some common uses are to allow a cdrom to be changed even if locked, or to answer the question as to whether a VM was copied or moved.
+            - Can be used if O(state=powered-on).
```
mikemorency, line 108:
```suggestion
            - Can be used if O(state) is V(powered-on).
```

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+            )
+            self.module.params['folder'] = fq_folder_path.rstrip('/')
+
+    def get_vm(self):
+        """
+        Finds and gets VM according to the specified name/uuid/moid
+        Returns:
+            list(vm), or None if no matches were found 
+        """
+        vm = self.get_vm_using_params(name_param=self.module.params['name'], moid_param=self.module.params['moid'], uuid_param=self.module.params['uuid'])
```
mikemorency, line 275:
```suggestion
        vm = self.get_vm_using_params(fail_on_missing=True)
```
The function defaults already work for you. The only thing you need is `fail_on_missing` since the module needs a VM.

FYI - The function params you originally supplied should just be the name of that param in the module inputs. So in your case `name_param='name'`, `moid_param='moid'`, `uuid_param='uuid'`

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+        # if self.module.params['name']:
+        #     vm = self.get_vm_using_params(name_param=self.module.params['name'])
+        # elif self.module.params['moid']:
+        #     vm = self.get_vm_using_params(moid_param=self.module.params['moid'])
+        # elif self.module.params['uuid']:
+        #     vm = self.get_vm_using_params(uuid_param=self.module.params['uuid'])
+
+        if not vm:
+            id = self.module.params.get('uuid') or self.module.params.get('moid') or self.module.params.get('name')
+            self.module.fail_json(msg="Unable to set power state for non-existing virtual machine : '%s'" % id)
```
mikemorency, line 287: This check is already handled by `self.get_vm_using_params(fail_on_missing=True)`

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+        return False
+    
+    def get_current_and_desired_states(self, vm):
+        state = self.module.params['state']
+        if state == 'present':
+            state = 'poweredon'
+        desired_state = state.replace('_', '').replace('-', '').lower()
+        current_state = vm.summary.runtime.powerState.lower()
+
+        return current_state, desired_state
```
mikemorency, line 568: you can definitely combine these two

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec={
+            **base_argument_spec(), **dict(
+                datacenter=dict(type='str', required=True),
+                state=dict(type='str', default='present',
+                        choices=['present', 'powered-off', 'powered-on', 'reboot-guest', 'restarted', 'shutdown-guest', 'suspended']),
```
mikemorency, line 577: you should remove `present` as an option here. It doesnt make sense. I see some other places in the module where this is referenced, so youll need to update there as well

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+                    if hasattr(task.info.error, 'thumbprint'):
+                        host_thumbprint = task.info.error.thumbprint
+                except AttributeError:
+                    pass
+                finally:
+                    raise_from(TaskError(error_msg, host_thumbprint), task.info.error)
+            if task.info.state in [vim.TaskInfo.State.running, vim.TaskInfo.State.queued]:
+                sleep_time = min(2 ** failure_counter + randint(1, 1000) / 1000, max_backoff)
+                time.sleep(sleep_time)
+                failure_counter += 1
```
mikemorency, line 363: This function can be removed. You can use the RunningTaskMonitor class here:
https://github.com/ansible-collections/vmware.vmware/blob/main/plugins/module_utils/_vmware_tasks.py#L34


You can see an example of how to use that class here: https://github.com/ansible-collections/vmware.vmware/blob/main/plugins/modules/esxi_maintenance_mode.py#L184-L195
In your module, `task = self.host.ExitMaintenanceMode_Task(self.module.params['timeout'])` would be equivalent to  `task = vm.PowerOff()` or `task = vm.PowerOn()` or any of the others like that.
Also you probably dont need to catch those `vmodl` errors

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+        while timeout > 0:
+            if vm.runtime.powerState.lower() == 'poweredoff':
+                break
+            time.sleep(interval)
+            timeout -= interval
+        else:
+            self.result['failed'] = True
+            self.result['msg'] = 'Timeout while waiting for VM power off.'
+
+    def set_vm_powerstate(self, vm, force, timeout=0, answers=None):
```
mikemorency, line 376: This should be broken up, 3 levels of if/else seems crazy and this has like 6. Also, you dont need the `force` parameter anymore, since you can just access `self.params['force']`. Maybe the same is for `answers`, I havnt looked

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,634 @@
...
+    def standardize_folder_param(self):
+        """
+        Removes all "/" characters in the folder path if the folder param exists
+        """
+        if self.module.params['folder']:
+            fq_folder_path = format_folder_path_as_vm_fq_path(
+                self.module.params['folder'],
+                self.module.params['datacenter']
+            )
+            self.module.params['folder'] = fq_folder_path.rstrip('/')
```
mikemorency, line 267: you dont need this line, i added it to the format function in a recent pr

plugins/modules/vmware_guest_powerstate.py
OrrBG: wdym combine?

plugins/modules/vmware_guest_powerstate.py
OrrBG: so you want there to be only powered-on and that the default value will be powered-on?

plugins/modules/vmware_guest_powerstate.py
mikemorency: like just have one method, instead of having `current_state_matches_desired_state` call `get_current_and_desired_states` basically
```
        desired_state = state.replace('_', '').replace('-', '').lower()
        current_state = vm.summary.runtime.powerState.lower()
        return desired_state == current_state
```

mikemorency: i see. so since these things (and `vm`) are key to the module functioning, you could put them in the init method:
```
def __init__(self, ....):
....
self.vm = self.get_vm_by_params()  # isnt the get_vm() method basically just a one liner too?
self.desired_state = state.replace('_', '').replace('-', '').lower()
self.current_state = self.vm.summary.runtime.powerState.lower()
```

then you probably dont need a `current_state_matches_desired_state` method, you can just do the equality check in main. and you can still reference the state variables later on

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,559 @@
...
+            if self.vm.runtime.powerState.lower() == 'poweredoff':
+                break
+            time.sleep(interval)
+            timeout -= interval
+        else:
+            self.result['failed'] = True
+            self.result['msg'] = 'Timeout while waiting for VM power off.'
+
+            
+    def fail_with_message(self, msg):
```
mikemorency, line 321: you should call `module.fail_json(msg="whatever")` instead of this method

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,559 @@
...
+    )
+
+    result = dict(
+        changed=False,
+        result={}
+    )
+
+    vmware_guest_powerstate = VmwareGuestPowerstateModule(module)
+    vmware_guest_powerstate.standardize_folder_param()
+    if vmware_guest_powerstate.current_state_matches_desired_state():
```
mikemorency, line 542:
```suggestion
    if vmware_guest_powerstate.current_state == vmware_guest_powerstate.desired_state:
```
this method is redundant

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,559 @@
...
+                    else:
+                        self.fail_with_message('Cannot suspend virtual machine in the current state %s' % self.current_state)
+
+                elif self.desired_state in ['shutdownguest', 'rebootguest']:
+                    if self.current_state == 'poweredon':
+                        if self.vm.guest.toolsRunningStatus == 'guestToolsRunning':
+                            if self.desired_state == 'shutdownguest':
+                                task = self.vm.ShutdownGuest()
+                                if timeout > 0:
+                                    self.wait_for_poweroff(timeout)
```
mikemorency, line 374: you should incorporate the timeout into `run_vm_configuration_task`. Theres no reason it only needs to apply to the `shutdownguest` state

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,559 @@
...
+                _, task_result = RunningTaskMonitor(task).wait_for_completion()
+            except TaskError as e:
+                self.fail_with_message(to_text(e))
+            finally:
+                if task.info.state == 'error':
+                    self.fail_with_message(task.info.error.msg)
+                else:
+                    self.result['changed'] = True
+
+    def set_vm_powerstate(self, force, timeout=0):
```
mikemorency, line 337: please break this into smaller methods

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,559 @@
...
+            ['scheduled_at', 'question_answers']
+        ],
+        required_one_of=[
+            ['name', 'uuid', 'moid']
+        ],
+    )
+
+    result = dict(
+        changed=False,
+        result={}
```
mikemorency, line 537: should this be removed?

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,497 @@
...
+        module.exit_json(**vmware_guest_powerstate.result)
+
+    if module.check_mode:
+        vmware_guest_powerstate.result['changed'] = True
+        module.exit_json(**vmware_guest_powerstate.result)
+    
+    vmware_guest_powerstate.configure_vm_powerstate()
+    
+    if hasattr(vmware_guest_powerstate.result, "failed") and vmware_guest_powerstate.result.get('failed') is True:
+        module.fail_json(**vmware_guest_powerstate.result)
```
mikemorency, line 491: i dont think you need this anymore, since you are using `fail_json` inside the VmwareGuestPowerstateModule methods

plugins/modules/vmware_guest_powerstate.py
```diff
@@ -0,0 +1,552 @@
...
+            if self.vm.runtime.powerState.lower() == 'poweredoff':
+                break
+            time.sleep(interval)
+            timeout -= interval
+        else:
+            self.result['failed'] = True
+            self.result['msg'] = 'Timeout while waiting for VM power off.'
+
+    def run_vm_configuration_task(self, task, timeout):
+        if task:
```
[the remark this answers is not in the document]
OrrBG: after some trial and error I found that this condition is crucial for interfacing with the API since the shutdown and reboot API functions return None while the rest of the powerstate functions return tasks, so everything that happens after this condition  (all the functions used in _vmware_tasks) will result in the program crashing

plugins/modules/vm_powerstate.py
```diff
@@ -0,0 +1,468 @@
...
+                else:
+                    task = self.vm.RebootGuest()
+                # Set result['changed'] immediately because
+                # shutdown and reboot return None.
+                self.result['changed'] = True
+            else:
+                self.module.fail_json(msg="VMware tools should be installed for guest shutdown/reboot")
+        elif self.current_state == 'poweredoff':
+            self.result['changed'] = False
+            return None
```
mikemorency, line 302: this seems odd to me. I think it should raise an error or something?

mikemorency: i guess it depends on desired state. if the user wanted to reboot then this should be an error, otherwise this looks right

plugins/modules/vm_powerstate.py
```diff
@@ -0,0 +1,468 @@
...
+    )
+
+    vm_powerstate = VmPowerstateModule(module)
+    vm_powerstate.standardize_folder_param()
+    # Check if a virtual machine is locked by a question
+    if (not module.check_mode) and vm_powerstate.vm.runtime.question:
+        if not vm_powerstate.module.params['question_answers']:
+            module.fail_json(msg="No answers provided for question %s, set answers using the question_answers option"
+                              % vm_powerstate.vm.runtime.question.text)     
+        vm_powerstate.answer_questions()
```
mikemorency, line 454: the check_mode check should be inside this if:
```
if vm_powerstate.vm.runtime.question:
  result['changed'] = true
  if not module.check_mode:
    answer_questions()
```
you could also move all of this into the `answer_questions` method, since that is a pretty simple method right now and that might make this less messy

plugins/modules/vmware_guest_powerstate.py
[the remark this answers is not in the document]
mikemorency: i see, i think this should be a guard clause instead:
```
if not task:
  return
```

plugins/modules/vm_powerstate.py
OrrBG: why should the existence of a vm.runtime.question result in 'changed' being turned to true?

mikemorency: The VM has a question that needs to be answered, and this module is going to answer it. So the VM will be in a different state (not power state necessarily) than before the module was run

mikemorency (approved):

plugins/modules/vm_powerstate.py
```diff
@@ -0,0 +1,470 @@
...
+    use_instance_uuid:
+        description:
+            - Whether to use the VMware instance UUID rather than the BIOS UUID.
+        default: false
+        type: bool
+    folder:
+        description:
+            - Destination folder, absolute or relative path to find an existing guest.
+            - Should be the full folder path, with or without the 'datacenter/vm/' prefix
+            - For example 'datacenter name/vm/path/to/folder' or 'path/to/folder'
```
anna-savina, line 64:
```suggestion
            - For example 'datacenter_name/vm/path/to/folder' or 'path/to/folder'
```

plugins/modules/vm_powerstate.py
```diff
@@ -0,0 +1,470 @@
...
+    def run_vm_configuration_task(self, task):
+        if not task:
+            return
+        try:
+            succeeded, task_result = RunningTaskMonitor(task).wait_for_completion(
+                vm=self.vm, timeout=self.module.params['timeout'], answers=self.module.params.get('question_answers', None))
+        except TaskError as e:
+            self.module.fail_json(msg=to_text(e))
+        finally:
+                self.result['changed'] = True
```
anna-savina, line 276:
```suggestion
            self.result['changed'] = True
```

plugins/modules/vm_powerstate.py
```diff
@@ -0,0 +1,470 @@
...
+        scheduled_at = self.module.params.get('scheduled_at', None)
+        if scheduled_at:
+            scheduled_task_spec = self.configure_scheduled_task_spec(scheduled_at)
+            self.configure_vm_scheduled_powerstate(scheduled_task_spec)
+        else:
+            self.set_vm_powerstate()
+
+    def configure_scheduled_task_spec(self, scheduled_at):
+        """
+        Reurns:
```
anna-savina, line 349:
```suggestion
        Returns:
```

plugins/modules/vm_powerstate.py
```diff
@@ -0,0 +1,470 @@
...
+
+- name: Wait for the virtual machine to shutdown
+  vmware.vmware.vm_powerstate:
+    hostname: "{{ vcenter_hostname }}"
+    username: "{{ vcenter_username }}"
+    password: "{{ vcenter_password }}"
+    datacenter: "{{ vm_datacenter }}"
+    name: "{{ guest_name }}"
+    state: shutdown-guest
+    timeout: 200
```
anna-savina, line 175: Description of `timeout` arg says that "Can not be used if O(state) is V(shutdown-guest) or V(reboot-guest)." Is there a mistake here or in the description?

OrrBG: Correct! My mistake

tests/integration/targets/vmware_vm_powerstate/tasks/eco-vcenter.yml
```diff
@@ -0,0 +1,353 @@
...
+        password: "{{ vcenter_password }}"
+        folder: "{{ vm_folder }}"
+        name: "{{ vm }}"
+        state: powered-off
+        scheduled_at: "{{ '%d/%m/%Y %H:%M' | strftime((ansible_date_time.epoch | int) + 86400)}}"
+        scheduled_task_name: "task_01"
+        scheduled_task_description: "Sample task to poweroff VM"
+        scheduled_task_enabled: true
+
+    - name: Wait until VM is ready
```
anna-savina, line 256: Should the VM be powered off after the prev task?

mikemorency: the previous task schedules a power off action for the future....myabe instead of 86400 seconds we should just do like 5 seconds, then wait 10 seconds and check if the vm was actually powered off

anna-savina (approved):
LGTM. Just update the branch and fix the unit tests please

mikemorency:
@OrrBG is this good to merge?

PR #126, opened 2025-02-25
migrate/vmware_guest_snapshot

OrrBG:
##### SUMMARY
migrate vmware_guest_snapshot


plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,497 @@
...
+        except Exception as e:
+            self.module.fail_json(msg=to_text(e))
+
+        if not task:
+            return
+
+        try:
+            success, task_result = RunningTaskMonitor(task).wait_for_completion(vm=self.vm)
+        except TaskError as e:
+            self.module.fail_json(msg=to_text(e))
```
mikemorency, line 444: combine these try/excepts

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,497 @@
...
+
+        self.result['changed'] = True
+        self.result['snapshot_results'] = self.list_snapshots(self.vm)
```
mikemorency, line 447: i dont think we want to return all snapshot data (unless i guess its a remove_all state). just the snapshot that was changed

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,497 @@
...
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec={
+            **base_argument_spec(), **dict(
+                state=dict(default='present', choices=['present', 'absent', 'rename', 'revert', 'remove_all']),
```
mikemorency, line 454: this module feels overloaded with functionality, but Im not sure if it should be broken up or just re-organized

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,497 @@
...
+        snap_obj = []
+        for snapshot in snapshots:
+            if snapshot.id == snapid:
+                snap_obj.append(snapshot)
+            else:
+                snap_obj = snap_obj + self.get_snapshots_by_id_recursively(snapshot.childSnapshotList, snapid)
+        return snap_obj
+
+    def get_snapshots_recursively(self):
+        if self.module.params["snapshot_name"]:
```
mikemorency, line 377: Anywhere you have `self.module.params` you can just use `self.params`

plugins/modules/vm_snapshot.py
OrrBG: I think all the functions  form a cohesive "unit", other than 'revert' which seems like it might be better as a separate module but then it might be weird to have a module with only one function. 'revert' does make sense as part of this module as a 'snapshot' functionality although all the other functions are 'taking a snapshot' functionality while 'revert' is a 'use snapshot' functionality. It makes sense to me that they would form one module but we could separate them if you think that that would adhere better to 'single responsibility principle'.

mikemorency: I agree with you that they all make sense to be in one module. After some more thought, I think my issue is that `present` and `absent` are states while `revert` and `remove_all` are actions. I am not sure I have a better alternative though, since `present` and `absent` are standard values for status....

OrrBG: lmk what you think of the changes

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,449 @@
...
+
+        try:
+            task = snapshot_state_function[self.params['state']]()
+            if not task:
+                return
+            success, task_result = RunningTaskMonitor(task).wait_for_completion(vm=self.vm)
+        except Exception as e:
+            self.module.fail_json(msg=to_text(e))
+
+        self.result['snapshot_results'] = self.list_snapshots()
```
mikemorency, line 394: Two more general points relating to this method:
1. Now that the state options are simpler, this method (and the "entry" into the module in general) seems overly complex. For example, its not clear to me why check_mode is showing something is changed at this point, or how the rename operations come into play
2. The module should not always list all the snapshots for the VM. That can be the _info modules job. This should just show the impacted snapshot(s). The task object will reference the snap object, so you dont need to re-collect all of the snaps again

plugins/modules/vm_snapshot.py
mikemorency: it flows a lot better now

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,449 @@
...
+            **base_argument_spec(), **dict(
+                state=dict(default='present', choices=['present', 'absent']),
+                name=dict(type='str'),
+                name_match=dict(type='str', choices=['first', 'last'], default='first'),
+                uuid=dict(type='str'),
+                moid=dict(type='str'),
+                use_instance_uuid=dict(type='bool', default=False),
+                remove_all=dict(type='bool', default=False),
+                folder=dict(type='str'),
+                datacenter=dict(required=True, type='str'),
```
mikemorency, line 409: see https://github.com/ansible-collections/community.vmware/pull/2300

we should implement that. I think its pretty easy at this point. I _think_ making folder and datacenter optional is all that would be required

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,449 @@
...
+            result['current_snapshot'] = self.serialize_snapshot_obj_to_json(current_snap_obj)
+        else:
+            result['current_snapshot'] = dict()
+        return result
+
+    def get_snapshots_by_identifier_recursively(self, snapshots, snapidentifier):
+        for snapshot in snapshots:
+            if ((isinstance(snapidentifier, vim.vm.Snapshot) and snapidentifier == snapshot.snapshot) or
+                (isinstance(snapidentifier, int) and snapidentifier == snapshot.id) or
+                (isinstance(snapidentifier, str) and snapidentifier == snapshot.name)):
```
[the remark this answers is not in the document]
OrrBG: I check for equality in snapshot object for when `list_snapshots` calls this function using the snapshot object. Checking id, name and snapshot is necessary because snapidentifier sometimes refers to the name, sometimes to the id and sometimes to the snapshot. And according to the vmware documentation id is an int
Screenshot from 2025-03-18 11-48-17

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,449 @@
...
+            return snapshot
+        except vim.fault.RestrictedVersion as e:
+            self.module.fail_json(msg="Failed to take snapshot due to VMware Licence"
+                                      " restriction : %s" % to_native(e.msg))
+        except Exception as e:
+            self.module.fail_json(msg="Failed to create snapshot of virtual machine"
+                                      " %s due to %s" % (self.params['name'], to_native(e)))
+
+    def rename_snapshot(self):
+        if self.params["new_snapshot_name"] and self.params["new_description"]:
```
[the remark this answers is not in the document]
OrrBG: Right but that comes at the cost of always updating both name and description when its entirely possible that we don't need to update anything and the user just wants to check that the vm is present. The overhead is probably pretty minor but the question is if its worth to execute an uncalled for job just in order to make this code look a little cleaner

plugins/modules/vm_snapshot.py
OrrBG: list_snapshot returns only the addressed snapshot

mikemorency: yes, your right. ignore that point

plugins/modules/vm_snapshot.py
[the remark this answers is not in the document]
mikemorency: Oh i see. Confusing that snapshotTree is an object but so is virtualMachineSnapshot.....

for `list_snapshot`, isnt `self.vm.snapshot.currentSnapshot` the affected snap? Do you need to call `get_snapshots_by_identifier_recursively` at all?

plugins/modules/vm_snapshot.py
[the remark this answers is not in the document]
mikemorency: I agree with you, and I think that check should happen before you call this method. For example

```
if self.params["new_snapshot_name"] and self.params["new_description"]:
  self.rename_snapshot()
.....
def rename_snapshot(self):
  task = self.snap_object.RenameSnapshot(
      name=(self.params["new_snapshot_name"] or self.snap_object.name),
      description=(self.params["new_description"] or self.snap_object.description)
  )
  ....
```

plugins/modules/vm_snapshot.py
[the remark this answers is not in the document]
mikemorency: also I dont think you need to check the indentifier instance. I dont see any issue with just checking:
`snapidentifier == snapshot.id or snapidentifier == snapshot.name`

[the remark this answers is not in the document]
OrrBG: according to the currentSnapshot documentation  "This property is set by calling [RevertToSnapshot_Task] or [CreateSnapshot_Task]." so it won't necessarily always refer to the snapshot we want

tests/integration/targets/vmware_vm_snapshot/tasks/main.yml
```diff
@@ -0,0 +1,182 @@
...
+        vcenter_hostname: "{{ vcenter_hostname }}"
+        vcenter_username: "{{ vcenter_username }}"
+        vcenter_password: "{{ vcenter_password }}"
+        vcenter_validate_certs: false
+        name: "{{ test_resource_pool }}"
+        parent: "{{ vcenter_rp_id }}"
+      register: _test_rp
+
+    - name: Deploy Virtual Machine from template in content library
+      community.vmware.vmware_content_deploy_template:
```
mikemorency, line 29: use vmware.vmware.deploy_content_library_ovf please. We should use vmware.vmware or vmware.vmware_rest modules whenever possible

mikemorency (approved):

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,416 @@
...
+description:
+    - This module can be used to create, delete and update snapshot(s) of the given virtual machine.
+author:
+    - Ansible Cloud Team (@ansible-collections)
+options:
+    state:
+        description:
+        - Manage snapshot(s) attached to a specific virtual machine.
+        - If set to V(present) and the snapshot absent, a new snapshot will be created with the given name.
+        - If set to V(present), the snapshot is present, and O(new_snapshot_name) or O(description) are set, the snaphots name or description is updated.
```
anna-savina, line 26:
```suggestion
        - If set to V(present), the snapshot is present, and O(new_snapshot_name) or O(description) are set, the snapshot name or description is updated.
```

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,416 @@
...
+                description=dict(type='str', default=''),
+                quiesce=dict(type='bool', default=False),
+                memory_dump=dict(type='bool', default=False),
+                remove_children=dict(type='bool', default=False),
+                new_snapshot_name=dict(type='str'),
+            )
+        },
+        supports_check_mode=True,
+        required_if=[
+            ['state', 'present', ('snapshot_name',)]
```
anna-savina, line 388:
```suggestion
            ('state', 'present', ('snapshot_name')),
            ('remove_all', False, ('snapshot_name', 'snapshot_id'), True)
```

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,416 @@
...
+        ],
+        required_together=[
+            ['name', 'folder']
+        ],
+        required_one_of=[
+            ['snapshot_name', 'snapshot_id', 'remove_all'],
```
anna-savina, line 394: remove_all has default value, even if it's False, doesn't that mean it's always specified?

OrrBG: good point

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,414 @@
...
+            ('remove_all', False, ('snapshot_name', 'snapshot_id'), True)
+        ],
+        required_together=[
+            ['name', 'folder']
+        ],
+        required_one_of=[
+            ['name', 'uuid', 'moid']
+        ],
+        mutually_exclusive=[
+            ['name', 'uuid', 'moid']
```
anna-savina, line 398:
```suggestion
            ('name', 'uuid', 'moid'),
            ('snapshot_name', 'snapshot_id')
```

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,416 @@
...
+        elif self.params["snapshot_name"]:
+            self.snap_object = self.get_snapshot_by_identifier_recursively(self.vm.snapshot.rootSnapshotList,
+                                                                           self.params["snapshot_name"])
+        elif self.params["snapshot_id"] and self.params['state'] == 'absent':
+            self.snap_object = self.get_snapshot_by_identifier_recursively(self.vm.snapshot.rootSnapshotList,
+                                                                           self.params["snapshot_id"])
+
+    def snapshot_vm(self):
+        if self.snap_object and (self.params['new_snapshot_name'] or self.params['description']):
+            return self.rename_snapshot()
```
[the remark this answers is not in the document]
anna-savina: and if a new_snapshot_name is the same as old one?
@mikemorency WDYT?

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,415 @@
...
+description:
+    - This module can be used to create, delete and update snapshot(s) of the given virtual machine.
+author:
+    - Ansible Cloud Team (@ansible-collections)
+options:
+    state:
+        description:
+        - Manage snapshot(s) attached to a specific virtual machine.
+        - If set to V(present) and the snapshot absent, a new snapshot will be created with the given name.
+        - If set to V(present), the snapshot is present, and O(new_snapshot_name) or O(description) are set, the snaphot name or description is updated.
```
anna-savina, line 26:
```suggestion
        - If set to V(present), the snapshot is present, and O(new_snapshot_name) or O(description) are set, the snapshot name or description is updated.
```

tests/integration/targets/vmware_vm_snapshot/tasks/main.yml
```diff
@@ -0,0 +1,150 @@
+---
+- block:
```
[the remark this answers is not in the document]
OrrBG: lmk what you think of the changes

plugins/modules/vm_snapshot.py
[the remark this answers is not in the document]
OrrBG: I think I take care of this in the new commit

tests/integration/targets/vmware_vm_snapshot/tasks/main.yml
```diff
@@ -0,0 +1,328 @@
...
+        password: "{{ vcenter_password }}"
+        datacenter: "{{ vm_datacenter }}"
+        folder: "/{{ vm_datacenter }}/vm/"
+        name: "{{ vm }}"
+        state: absent
+        snapshot_name: snap1
+        validate_certs: false
+
+    - name: Gather info from vm
+      community.vmware.vmware_guest_info:
```
anna-savina, line 63: Can be used `vmware.vmware.guest_info` instead of community module?

tests/integration/targets/vmware_vm_snapshot/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  - size_gb: 10
+    type: thin
+    autoselect_datastore: true
+vm_hardware:
+  memory_mb: 2000
+  num_cpus: 2
+  boot_firmware: efi
+
+datastore: "{{ shared_storage_02 }}"
+library: "{{ tiny_prefix }}-content-library"
```
anna-savina, line 18: `library` and `test_resource_pool` are not used vars

plugins/modules/vm_snapshot.py
```diff
@@ -0,0 +1,410 @@
...
+        type: str
+    snapshot_name:
+        description:
+        - The name of the snapshot to manage.
+        - Either this parameter or O(snapshot_id) is required if O(remove_all) is set to False.
+        type: str
+    snapshot_id:
+        description:
+        - The ID of the snapshot to manage. This option cannot be used when creating a new snapshot.
+        - Either this parameter or O(snapshot_id) is required if O(remove_all) is set to False.
```
anna-savina, line 81:
```suggestion
        - Either this parameter or O(snapshot_name) is required if O(remove_all) is set to False.
```

tests/integration/targets/vmware_vm_snapshot/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
...
+vm_folder: "/{{ vcenter_datacenter }}/vm/e2e-qe"
+vm_guest_id: "rhel8_64Guest"
+vm_disk:
+  - size_gb: 10
+    type: thin
+    autoselect_datastore: true
+vm_hardware:
+  memory_mb: 2000
+  num_cpus: 2
+  boot_firmware: efi
```
anna-savina, line 15: `vm_folder`, `vm_guest_id`, `vm_disk`, `vm_hardware` are not in use

tests/integration/targets/vmware_vm_snapshot/run.yml
```diff
@@ -0,0 +1,15 @@
...
+  collections:
+    - community.general
+
+  tasks:
+    - name: Import eco-vcenter credentials
+      ansible.builtin.include_vars:
+        file: ../../integration_config.yml
+      tags: eco-vcenter-ci
+
+    - name: Import vmware_vm_snapshot role
```
anna-savina, line 11:
```suggestion
    - name: Import vmware_vm_snapshot test
```

anna-savina:
@mikemorency should this module be added to the action group?

mikemorency:
Yes, good catch anna. @OrrBG can you add this module to `runtime/meta.yml` under `action_groups`? We can add it to the query file in a separate PR

anna-savina (approved):

PR #137, opened 2025-03-11
missed adding host filtering to vm inventory

mikemorency:
##### SUMMARY
The VM inventory is missing the actual method call to filter hosts out. This adds that in. There is another changelog that already captures this addition

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vms inventory

### DIFF
--- a/plugins/inventory/vms.py
+++ b/plugins/inventory/vms.py
@@ -279,6 +279,8 @@ class InventoryModule(VmwareInventoryBase):
         return hostvars
 
     def __update_inventory(self, vm):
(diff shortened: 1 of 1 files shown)


OrrBG:
looks good, seems like a changelog is needed

anna-savina (approved):

mikemorency:
> looks good, seems like a changelog is needed

Adding another bugfix changelog for features that havnt been released yet can just be confusing so we havnt added one historically. I think vmware_rest has a label you can add to skip that check but it hasnt been added here yet

PR #138, opened 2025-03-19
adding module to upload ovf to content library

mikemorency:
##### SUMMARY
Adding a module to upload an OVF or OVA file from a remote source to a content library. The remote source can be the Ansible host filesystem, an HTTPS/HTTP URL, a file URL, or a datastore URL.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
content_library_upload_ovf

##### ADDITIONAL INFORMATION
Originally, this was going to be a migration of the community.vmware.deploy_ovf module. That allows a user to deploy an OVF from a remote source. However, that module is complex and has a lot of extra features that would be more difficult to maintain.
I decided to create a module supports existing use cases in vmware_ops and vmware_rest and try to decrease the scope.

FYI @mariolenz  at this point, this module cannot replace the `deploy_ovf` module and there I currently have no plan to migrate more functionality. But I do think many `deploy_ovf` users could achieve the same thing using this module and 
`vmware.vmware.deploy_content_library_ovf`

### DIFF
--- a/plugins/modules/import_content_library_ovf.py
+++ b/plugins/modules/import_content_library_ovf.py
@@ -203,24 +203,23 @@ class VmwareRemoteOvf(ModuleRestBase):
         If a library name was supplied, search using that name.
         Otherwise, search with a name of None, which essentially means any name
 
-        Returns: list(str), list of IDs
+        Returns: str
         """
         if self._library_id:
-            return self._library_id
-
-        if self.params['library_id']:
+            pass
+        elif self.params['library_id']:
             self._library_id = self.params['library_id']
+        else:
+            self._library_id = self.get_content_library_ids(name=self.params['library_name'], library_type='LOCAL', fail_on_missing=True)[0]
 
-        self._library_id = self.get_content_library_ids(name=self.params['library_name'], library_type='LOCAL', fail_on_missing=True)[0]
         return self._library_id
 
     @property
     def library_item_id(self):
-        if self._library_item_id:
-            return self._library_item_id
+        if not self._library_item_id:
+            library_item_ids = self.get_library_item_ids(name=self.params['dest'], library_id=self.library_id)
+            self._library_item_id = library_item_ids[0] if library_item_ids else None
 
-        library_item_ids = self.get_library_item_ids(name=self.params['dest'], library_id=self.library_id)
-        self._library_item_id = library_item_ids[0] if library_item_ids else None
         return self._library_item_id
 
     def create_library_item(self):
--- a/plugins/modules/import_content_library_ovf.py
+++ b/plugins/modules/import_content_library_ovf.py
@@ -1,0 +1,499 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2023, Ansible Cloud Team (@ansible-collections)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: import_content_library_ovf
+short_description: Import an OVF or OVA file to a content library from a remote source.
+description:
+    - Import an OVF or OVA library item from a remote source, such as a file on the Ansible host or a URL.
+    - Source files should be OVAs or directories matching the OVF standard. OVF directories should contain an .ovf
+      file and at least on .vmdk file.
+    - This module manages the content library item, not the remote OVF/OVA. If state is absent, only the content library item
+      will be affected.
+
+author:
+    - Ansible Cloud Team (@ansible-collections)
+
+seealso:
+    - module: vmware.vmware.deploy_content_library_ovf
+
+extends_documentation_fragment:
+    - vmware.vmware.base_options
+    - vmware.vmware.additional_rest_options
+
+options:
+    src:
+        description:
+            - The source OVF or OVA file that should be imported to the content library. This can be a local path or a URL.
+            - If it is a local path, it should be a valid path on the Ansible host.
+            - If it is a URL, its scheme can be HTTPS, HTTP, or datastore (ds://). The file is acquired directly
+              from the vCenter appliance.
+            - The module will not encode URLs for you. If this is a URL with special characters (like $ or ~), you should
+              use the urlencode filter.
+            - Required when state is present.
+        type: path
+        required: false
+        aliases: [url, path]
+    dest:
+        description:
+            - The destination name of the OVF item in the content library.
+            - If an item with the same name already exists, the module will not update it.
+        type: str
+        required: true
+        aliases: [name]
+    state:
+        description:
+            - Controls if the OVF file should be present or absent in the content library.
+        type: str
+        choices: [present, absent]
+        default: present
+    description:
+        description:
+            - A description for the OVF item in the content library.
+        type: str
+        required: false
+    library_id:
+        description:
+            - The ID of the library to search within.
+            - One of O(library_id) or O(library_name) must be provided.
+        type: str
+        required: false
+    library_name:
+        description:
+            - The name of the library to search within.
+            - One of O(library_id) or O(library_name) must be provided.
+        type: str
+        required: false
+    ssl_thumbprint:
+        description:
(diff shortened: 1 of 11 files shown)


plugins/modules/upload_content_library_ovf.py
```diff
@@ -0,0 +1,487 @@
...
+    checksum_algorithm:
+        description:
+            - The checksum algorithm to use when validating uploads.
+            - This is required if O(checksum) is provided.
+        type: str
+        required: false
+        choices: ['SHA1', 'MD5', 'SHA512', 'SHA256']
+    checksum:
+        description:
+            - The checksum that should be used to validated the upload.
```
anna-savina, line 90:
```suggestion
            - The checksum that should be used to validate the upload.
```

plugins/modules/upload_content_library_ovf.py
```diff
@@ -0,0 +1,487 @@
...
+        try:
+            self.__start_upload(file_map=file_map)
+            self.__handle_warnings()
+            self.__wait_for_upload()
+        except Exception as e:
+            self.module.fail_json(msg="Failed to complete upload of OVF to vCenter: %s" % to_native(e))
+        finally:
+            self.__cleanup_tranfer()
+
+    def __cleanup_tranfer(self):
```
anna-savina, line 357:
```suggestion
    def __cleanup_transfer(self):
```
and in a line above

tests/integration/targets/vmware_upload_content_library_ovf/tasks/eco-vcenter.yml
```diff
@@ -0,0 +1,89 @@
...
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        validate_certs: false
+        src: "{{ test_src_url }}"
+        dest: "{{ test_dest }}"
+        library_name: "{{ ci_resources_content_library }}"
+      register: _test_url_idem
+
+    - name: Assert
```
anna-savina, line 30:
```suggestion
    - name: Assert upload
```

tests/integration/targets/vmware_upload_content_library_ovf/tasks/eco-vcenter.yml
```diff
@@ -0,0 +1,89 @@
...
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        validate_certs: false
+        dest: "{{ test_dest }}"
+        library_name: "{{ ci_resources_content_library }}"
+        state: absent
+      register: _test_url_del
+
+    - name: Assert
```
anna-savina, line 48:
```suggestion
    - name: Assert deletion
```

tests/integration/targets/vmware_upload_content_library_ovf/tasks/eco-vcenter.yml
```diff
@@ -0,0 +1,89 @@
...
+        src: "{{ test_src_url }}"
+        dest: "{{ test_dest }}"
+        library_name: "{{ ci_resources_content_library }}"
+      register: _test_url_idem
+
+    - name: Assert
+      ansible.builtin.assert:
+        that:
+          - _test_url is changed
+          - _test_url.library_item.id == _test_url_idem.library_item.id
```
anna-savina, line 34: Check test please, this assertion failed:
```
fatal: [localhost]: FAILED! => {
    "assertion": "_test_url.library_item.id == _test_url_idem.library_item.id",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

plugins/modules/upload_content_library_ovf.py
```diff
@@ -0,0 +1,487 @@
...
+                }
+                open_url(
+                    url=file_info.upload_endpoint.uri, data=local_file,
+                    headers=headers, validate_certs=self.params['validate_certs']
+                )
+
+    def __handle_warnings(self):
+        """
+        Depending on module parameters, handle any warnings that have occurred during the file transfer.
+        If there are warnings and the user has disabled fail_on_warnigns, the warnings are essentially
```
anna-savina, line 319:
```suggestion
        If there are warnings and the user has disabled fail_on_warnings, the warnings are essentially
```

tests/integration/targets/vmware_upload_content_library_ovf/tasks/eco-vcenter.yml
mikemorency: I fixed this and a few other things I found after, but now i am pretty sure the vcenter is having an issue accepting API imports....Ill follow up with tzach

mariolenz:
> FYI @mariolenz at this point, this module cannot replace the `deploy_ovf` module

As far as I understand this is about uploading an OFV / OVA to a content library. It's pretty clear that this is not a replacement for `community.vmware.vmware_deploy_ovf` because this module doesn't have anything to do with a content library:

https://github.com/ansible-collections/community.vmware/blob/6966997df7cd987a7b5a5c867c95333480e919eb/plugins/modules/vmware_deploy_ovf.py#L17-L20

> But I do think many `deploy_ovf` users could achieve the same thing using this module and `vmware.vmware.deploy_content_library_ovf`

I agree. But there might be situations or good reasons why people don't want to / can't use a content library. So I don't think I'll deprecate `community.vmware.vmware_deploy_ovf`.

> I currently have no plan to migrate more functionality

I agree. Lets keep this focused on uploading an OFV / OVA to a content library, _not_ also deploy one. I'm with you there.

cc @ihumster FYI

plugins/modules/import_content_library_ovf.py
```diff
@@ -0,0 +1,476 @@
...
+            self.api_client.content.library.Item.delete(self.library_item_id)
+        except Exception as err:
+            self.module.fail_json(msg="%s" % to_native(err))
+
+    def state_absent(self, result):
+        if not self.library_item_id:
+            return
+        result['changed'] = True
+        result['library_item']['id'] = self.library_item_id
+        if not self.module.check_mode:
```
ihumster, line 406: Since `supports_check_mode=True`, add text messages about changes to the ansible console.

plugins/modules/import_content_library_ovf.py
```diff
@@ -0,0 +1,476 @@
...
+
+    def state_present(self, result):
+        if self.library_item_id:
+            result['library_item']['id'] = self.library_item_id
+            return
+
+        result['changed'] = True
+        file_map = self.get_file_map()
+        if self.module.check_mode:
+            return
```
ihumster, line 417: Since `supports_check_mode=True`, add text messages about changes to the ansible console.

ihumster (changes requested):
LGFM, Messages are needed for check_mode and can be approved.

plugins/modules/import_content_library_ovf.py
mariolenz: As far as I understand, a module should work just the same with and without check mode enabled. Apart from changing anything, that is.

anna-savina (approved):

plugins/modules/import_content_library_ovf.py
```diff
@@ -0,0 +1,499 @@
...
+        If a library name was supplied, search using that name.
+        Otherwise, search with a name of None, which essentially means any name
+
+        Returns: list(str), list of IDs
+        """
+        if self._library_id:
+            return self._library_id
+
+        if self.params['library_id']:
+            self._library_id = self.params['library_id']
```
OrrBG, line 212: wouldn't you want to return here and avoid calling `get_content_library_ids`

OrrBG: or better yet just have this all be in an "if elif" clause and in the end just call `return self._library_id` instead of returning the same thing in 3 different instances

plugins/modules/import_content_library_ovf.py
```diff
@@ -0,0 +1,499 @@
...
+
+        self._library_id = self.get_content_library_ids(name=self.params['library_name'], library_type='LOCAL', fail_on_missing=True)[0]
+        return self._library_id
+
+    @property
+    def library_item_id(self):
+        if self._library_item_id:
+            return self._library_item_id
```
OrrBG, line 220: you can avoid this line if you use a guard clause instead and in the end always return self._library_id

plugins/modules/import_content_library_ovf.py
mikemorency: whoops, yea that is a good catch

Issue #56, opened 2024-07-31
Drop Support for pyVmomi < 8.0.3 and vSAN Management SDK


mariolenz:
##### SUMMARY
Drop Support for pyVmomi < 8.0.3 and vSAN Management SDK for Python:

> Starting with vSphere 8.0 Update 3 release version, the vSAN Management SDK for Python is integrated into the Python SDK for the VMware vSphere API ([pyVmomi](https://developer.broadcom.com/sdks/pyvmomi/latest)). From the Python Package Index ([PyPi](https://pypi.org/project/pyvmomi/)), you can download a single package to manage vSAN, vCenter, and ESXi.

[vSAN Management SDK for Python](https://developer.broadcom.com/sdks/vsan-management-sdk-for-python/latest/)

##### ISSUE TYPE
- Documentation Report

##### COMPONENT NAME
requirements.txt
README.md

mariolenz:
> I suggest using the same file as the referred git repo. [vsphere-automation-sdk-python requirements.txt](https://github.com/vmware/vsphere-automation-sdk-python/blob/master/requirements.txt).

I think there are more dependencies in this file than this collection really needs. I wouldn't do this. I think this should be enough:

```
pyVmomi>=8.0.3.0.1
vmware-vcenter
vmware-vapi-common-client
```

> If you allow me, I can make a PR improve this.

The problem is that changing the `pyVmomi` requirement from `>=6.7.1` to `>=8.0.3.0.1` might be considered a breaking change by some people. Especially since `vmware-vcenter` and `vmware-vapi-common-client` require Python 3.8+. I don't know, but it's possible that some people are using this collection with older Python version. That's why I tend to do this in the next major version, not now.

Can't you just ignore the `requirements.txt` for now? AFAIK galaxy doesn't use, anyway.

mikemorency:
> Especially since vmware-vcenter and vmware-vapi-common-client require Python 3.8+

We still need to support ansible 2.16 which allows for python2.7 on the ansible controller. Since many people may run the VMware module from localhost, I cannot change the vsphere-automation-sdk-python dependency from the git URL to the pypi package yet

mariolenz:
> We still need to support ansible 2.16 which allows for python2.7 on the ansible controller.

[...]

PR #145, opened 2025-03-27
update pyvmomi min to 8.0.3.0.1

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/56
Updates the minimum pyvmomi sdk package version to 8.0.3.0.1

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pyVmomi>=6.7.1
+pyVmomi>=8.0.3.0.1
 git+https://github.com/vmware/vsphere-automation-sdk-python.git ; python_version >= '2.7'  # Python 2.6 is not supported
 aiohttp
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 pyVmomi>=8.0.3.0.1
-git+https://github.com/vmware/vsphere-automation-sdk-python.git ; python_version >= '2.7'  # Python 2.6 is not supported
+vmware-vcenter
+vmware-vapi-common-client
 aiohttp
--- a/tests/unit/requirements.txt
+++ b/tests/unit/requirements.txt
@@ -1,5 +1,5 @@
 requests
 pyVim
-pyvmomi
+pyVmomi>=8.0.3.0.1
 git+https://github.com/vmware/vsphere-automation-sdk-python.git
 aiohttp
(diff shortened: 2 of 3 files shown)


mikemorency:
reminder: i need to fix the changelog file name

requirements.txt
```diff
@@ -1,3 +1,3 @@
-pyVmomi>=6.7.1
+pyVmomi>=8.0.3.0.1
 git+https://github.com/vmware/vsphere-automation-sdk-python.git ; python_version >= '2.7'  # Python 2.6 is not supported
```
mariolenz, line 2:
```suggestion
pyVmomi>=8.0.3.0.1
vmware-vcenter
vmware-vapi-common-client
```

tests/unit/requirements.txt
```diff
@@ -1,5 +1,5 @@
 requests
 pyVim
-pyvmomi
+pyVmomi>=8.0.3.0.1
 git+https://github.com/vmware/vsphere-automation-sdk-python.git
```
mariolenz, line 4:
```suggestion
pyVmomi>=8.0.3.0.1
vmware-vcenter
vmware-vapi-common-client
```

mariolenz (commented):
> We still need to support ansible 2.16 which allows for python2.7 on the ansible controller. Since many people may run the VMware module from localhost, I cannot change the vsphere-automation-sdk-python dependency from the git URL to the pypi package yet

According to the [ansible-core support matrix](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix), 2.16 doesn't. Only on the target. Actually, it looks like the support for Python 2.7 on the controller has been dropped 3 1/2 years ago in 2.12. 

So I think it should be safe to replace the dependency on the GitHub repo with dependencies on `vmware-vcenter` and `vmware-vapi-common-client` PyPI packages.

I'm not 100% sure there, but I think you should have another look at this.

mikemorency:
great catch! i was looking at the wrong column. Ill take another look at using the pypi packages but they should be OK to include

mikemorency:
the test failure seems unrelated (another pr hit the same thing). Im going to wait to merge this until I can confirm the issue though

Issue #144, opened 2025-03-26
Make some module_utils public


mikemorency:
##### SUMMARY

Now that the collection has reached a steady state (more or less), we can make some module utils public so the community.vmware collection can use them. Not all modules are ready for this change.

Included utils:
- pymomi client
- rest client
- client errors
- argument_spec

mikemorency:
@mariolenz FYI. please let me know if there are others you want to be made public.
The pyvmomi module base can probably go public but I enjoy the flexability of it being private if your not going to use it right away.
Similar sentiment for the rest module base, although that could actually use some cleanup still

mariolenz:
I've started to work on re-using this code: ansible-collections/community.vmware#2369

I'm not sure when I can give you feedback on this. I hope using your pymomi client and client errors should work for me (should be simply replacing ansible-collections/community.vmware#2349 with using your code) , but I didn't have a closer look at rest client and argument_spec yet.

PR #147, opened 2025-03-27
make client and arg spec utils public

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/144

Makes the client and argument module utils public, so other collections (namely community.vmware) can use them. Although this means we cannot change the utils as freely as we have been, they should be in a stable place and now we can synchronize with the community collection.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils/clients/pyvmomi
module_utils/clients/rest
module_utils/clients/errors
module_utils/argument_spec


mariolenz:
Cool! Thanks for this!

mariolenz:
[This](https://github.com/ansible-collections/community.vmware/pull/2369#issuecomment-2759102354) looks very promising :smiley:

anna-savina (approved):

mariolenz:
@anna-savina FYI I think that it should work to re-use this in `community.vmware`, but I wonder why this [fails in the CI](https://github.com/ansible-collections/community.vmware/pull/2369#issuecomment-2763655881) :-/

PR #150, opened 2025-04-01
add vm search params to content_template

mikemorency:
##### SUMMARY
Adds standard VM search params to the content_template module (uuid, moid, name_match etc).
Also makes the module arg validation more accurate

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
content_template

##### ADDITIONAL INFORMATION
All of the changes are backwards compatible

### DIFF
--- a/plugins/modules/content_template.py
+++ b/plugins/modules/content_template.py
@@ -15,70 +15,112 @@ DOCUMENTATION = r'''
 module: content_template
 short_description: Manage template in content library from virtual machine.
 description:
-- Module to manage template in content library from virtual machine.
-- Content Library feature is introduced in vSphere 6.0 version.
-- This module does not work with vSphere version older than 67U2.
+    - Module to manage template in content library from virtual machine.
+    - Content Library feature is introduced in vSphere 6.0 version.
+    - This module does not work with vSphere version older than 67U2.
 author:
-- Ansible Cloud Team (@ansible-collections)
+    - Ansible Cloud Team (@ansible-collections)
 requirements:
-- vSphere Automation SDK
+    - vSphere Automation SDK
 options:
-    template:
-        description:
-        - The name of template to manage.
-        type: str
-        required: true
-    library:
-        description:
-        - The name of the content library where the template will be created.
-        type: str
-        required: true
     vm_name:
         description:
-        - The name of the VM to be used to create template.
-        - This attribute is required only when the state attribute is set to present
+            - The name of the VM to be used to create template.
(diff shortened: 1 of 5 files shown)


plugins/modules/content_template.py
```diff
@@ -15,70 +15,112 @@
...
+    vm_name_match:
+        description:
+            - If multiple virtual machines matching the name, use the first or last found.
+        default: first
+        choices: [ first, last ]
+        type: str
+    vm_uuid:
+        description:
+            - UUID of the instance to manage if known, this is VMware's unique identifier.
+            - This is required if O(vm_moid) or O(vm_uuid) is not supplied.
```
anna-savina, line 42:
```suggestion
            - This is required if O(vm_moid) or O(vm_name) is not supplied.
```

plugins/modules/content_template.py
```diff
@@ -15,70 +15,112 @@
...
+            - A VM identifier is required when O(state) is present.
+        type: str
+    vm_moid:
+        description:
+            - Managed Object ID of the instance to manage if known, this is a unique identifier only within a single vCenter instance.
+            - This is required if O(vm_moid) or O(vm_uuid) is not supplied.
```
anna-savina, line 48:
```suggestion
            - This is required if O(vm_name) or O(vm_uuid) is not supplied.
```

plugins/modules/content_template.py
```diff
@@ -15,70 +15,112 @@
...
+        required: true
     host:
         description:
-        - Host onto which the virtual machine template should be placed.
-        - If O(host) and O(resource_pool) are both specified, O(resource_pool)
-          must belong to O(host).
-        - If O(host) and O(cluster) are both specified, O(host) must be a member of O(cluster).
-        - This attribute was added in vSphere API 6.8.
+            - Host onto which the virtual machine template should be placed.
+            - If O(host) and O(resource_pool) are both specified, O(resource_pool)
```
anna-savina, line 85: Is `O(resource_pool) must belong to O(host)` right?

anna-savina: I meant is this statement correct? The resource pool belongs to the cluster, not the host. Or am I missing something?

mikemorency: You can have standalone hosts that have resource pools....but I guess the user could specify a host in the cluster and then I am not sure any pool "belongs" to the host.

anna-savina (approved):

PR #151, opened 2025-04-01
fix license info test

mikemorency:
##### SUMMARY
Fixes the license info test. Someone added a new license to the lab, so the old check is invalid. I think it makes sense to just check if more than one license is returned, since we know that should always be true and the exact number isnt as important

##### ISSUE TYPE
- Bugfix Pull Request


anna-savina (approved):

anna-savina:
Test on simulator failed. Previously it was `licenses_length: 0` and `__res.licenses | length == licenses_length`
Maybe no reason to run this test on simulator at all?

PR #160, opened 2025-04-09
add cluster module queries

mikemorency:
##### SUMMARY
Adds any missing queries and cluster related queries for indirect node counting.
Also fixes a missing module in the action groups.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
cluster modules
event query file


mikemorency:
Adding the cluster output to `cluster_vcls` showed that the unit tests were not doing much of anything. When I updated the tests, it showed a potential bug if a user set only one of `datastores_to_add` or `datastores_to_remove` because there were no default values.

anna-savina (approved):

Issue #161, opened 2025-04-11
cluster_ha: Weird behaviour when replacing community.vmware.vmware_cluster_ha


mariolenz:
##### SUMMARY
`community.vmware.vmware_cluster_ha` is deprecated and I want to replace it with `vmware.vmware.cluster_ha`:

```
vmware.vmware.cluster_ha:
  storage_apd_response:
    mode: "warning"
  storage_pdl_response_mode: "warning"
  enable: true
  host_failure_response:
    default_vm_restart_priority: "medium"
	restart_vms: true
  vm_monitoring:
    mode: "vmMonitoringDisabled"
  host_isolation_response: "none"
  admission_control_policy: "cluster_resource"
  admission_control_failover_level: 1
```

But this fails.

~Fun fact: When I manually disable HA and run the task, the module enables HA again and _from then on it works_ 🙃~

~So I have a workaround to use the new module. Still, the module shouldn't explode like this IMHO.~

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cluster_ha

##### ANSIBLE VERSION
```
ansible [core 2.18.4]
  config file = None
  configured module search path = ['/home/lenzm02/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /data/venv-ansible-vmware/lib64/python3.12/site-packages/ansible
  ansible collection location = /home/lenzm02/.ansible/collections:/usr/share/ansible/collections
  executable location = /data/venv-ansible-vmware/bin/ansible
  python version = 3.12.8 (main, Dec  9 2024, 15:25:01) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)] (/data/venv-ansible-vmware/bin/python3.12)
  jinja version = 3.1.6
  libyaml = True
```

##### COLLECTION VERSION
```
# /data/venv-ansible-vmware/lib/python3.12/site-packages/ansible_collections
Collection    Version
------------- -------
vmware.vmware 1.10.1

# /data/venv-ansible-vmware/lib64/python3.12/site-packages/ansible_collections
Collection    Version
------------- -------
vmware.vmware 1.10.1
```

##### CONFIGURATION
```
CONFIG_FILE() = None

GALAXY_SERVERS:
```

##### OS / ENVIRONMENT
RHEL 8 + vSphere 8.0 U3 although I don't think that's relevant.

##### STEPS TO REPRODUCE
[...]

PR #162, opened 2025-04-13
fix cluster ha diff

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/161

When comparing VM config diffs, there is a syntax issue

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cluster_ha


anna-savina (approved):

plugins/modules/cluster_ha.py
```diff
@@ -431,7 +431,7 @@ def __check_vm_monitoring_config_diff(self, ha_config):
         if (
             vm_config.vmMonitoring != vm_params["mode"] or
             vm_config.failureInterval != vm_params["failure_interval"] or
-            vm_config.minUpTime != vm_params.get["minimum_uptime"] or
+            vm_config.minUpTime != vm_params.get("minimum_uptime") or
```
mariolenz, line 434: Why do you use `vm_params.get()` here but `vm_params[]` in the other tests? Wouldn't it be better to use one _or_ the other consistently?

mikemorency: yea. I kinda assumed it was there for a reason, but looking at the rest of the docs/defaults, it is not.

mariolenz (approved):
LGTM

I tried again with this change applied and the module worked fine. Thanks @mikemorency!

Issue #30, opened 2024-05-28
Remove deprecated module vm_list_group_by_clusters


mariolenz:
##### SUMMARY
Remove deprecated module `vm_list_group_by_clusters`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
vm_list_group_by_clusters

##### ADDITIONAL INFORMATION
#25

mikemorency:
closing as completed

PR #169, opened 2025-04-17
release 2.0.0

mikemorency:
##### SUMMARY
Release 2.0.0

##### ISSUE TYPE

- Release Pull Request

##### ADDITIONAL INFORMATION
Includes the tombstoning of vm_list_group_by_clusters, since that could not be done in an earlier pull request: https://github.com/ansible-collections/vmware.vmware/issues/30

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,50 @@ vmware.vmware Release Notes
 
 .. contents:: Topics
 
+v2.0.0
+======
+
+Major Changes
+-------------
+
+- cluster modules - Add identifying information about the cluster managed to the output of cluster modules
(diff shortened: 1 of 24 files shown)


OrrBG (approved):

anna-savina (approved):

mariolenz (changes requested):
I think you could simplify

https://github.com/ansible-collections/vmware.vmware/blob/50de4f69b4df169201575264f3aee368d97b2b37/plugins/modules/vm_list_group_by_clusters_info.py#L198-L203

to have only the `module.exit_json()` from the else branch.

I think this is the only place where the module needs a change, but it might help if someone else has another look.

PR #188, opened 2025-05-21
Update integration tests to use ansible 2.19

mikemorency:
This updates the integration tests to use ansible 2.19. We have to use a release candidate for now but will go back to the latest stable version when 2.19 is officially released


mariolenz:
I've only had a very quick look at this, but is it possible that the errors are due to problems with `vmware.vmware_rest`? I see a lot of errors with tasks named `Create a test resource pool` which seem to use a module from this collection.

mariolenz:
I wonder if we can use `Depends-On` here. It works in community.general, but I'm using Zuul there so I'm not sure...

tests/integration/requirements.txt
```diff
@@ -1,4 +1,4 @@
-ansible-core>=2.18.0,<2.19.0
+ansible-core==2.19.0b4
```
mariolenz, line 1:
```suggestion
ansible-core==2.19.0b5
```

Just to make sure, lets test with 2.19.0b5 which has been released recently.

mariolenz: I've changed this from 2.19.0b4 to 2.19.0b5.

mariolenz:
Hooray, all checks have passed :partying_face: 

@mikemorency I think you should mention this in #171. I hope this is enough to close the issue.

mariolenz (approved):
LGTM

Thanks @mikemorency!

Issue #190, opened 2025-05-27
vmware.vmware.cluster_ha: advanced_settings not working


pandalec:
##### SUMMARY

Previous ha advanced settings could be set with

```yaml
        advanced_settings:
          das.ignoreInsufficientHbDatastore: "true"
```

but with vmware.vmware.cluster_ha it results in:

```
('The request refers to an unexpected or unknown type.', None)
```

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
ansible [core 2.17.12]
```

##### COLLECTION VERSION

```paste below
vmware.vmware             2.0.1
```

##### CONFIGURATION

```paste below
ANY_ERRORS_FATAL(./ansible.cfg) = True
CALLBACKS_ENABLED(./ansible.cfg) = ['profile_tasks']
CONFIG_FILE() = ./ansible.cfg
DEFAULT_FORKS(./ansible.cfg) = 50
DEFAULT_JINJA2_NATIVE(./ansible.cfg) = True
DEFAULT_LOAD_CALLBACK_PLUGINS(./ansible.cfg) = True
DEFAULT_STDOUT_CALLBACK(./ansible.cfg) = debug
DISPLAY_SKIPPED_HOSTS(./ansible.cfg) = False
EDITOR(env: EDITOR) = hx
HOST_KEY_CHECKING(./ansible.cfg) = True
```

##### OS / ENVIRONMENT
Arch Linux

##### STEPS TO REPRODUCE

Change from 
```yaml
    - name: Enable and configure HA
      community.vmware.vmware_cluster_ha:
        advanced_settings:
          das.ignoreInsufficientHbDatastore: "true"
          das.usedefaultisolationaddress: "false"
```
To:

```yaml
    - name: Enable and configure HA
      vmware.vmware.cluster_ha:
        advanced_settings:
          das.ignoreInsufficientHbDatastore: "true"
          das.usedefaultisolationaddress: "false"
```

I also tried:

```yaml
"das.ignoreInsufficientHbDatastore": "true"
das.ignoreInsufficientHbDatastore: true
```

##### EXPECTED RESULTS

No error in advanced_settings

##### ACTUAL RESULTS

```paste below
('The request refers to an unexpected or unknown type.', None)
```

PR #192, opened 2025-06-02
cluster ha advanced setting bugfix

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/190

When a user tries to specify a non-string value for cluster HA or DRS advanced settings, an error is thrown.
This error is vague and unhelpful for the HA module. DRS actually has a decent error, so I didnt modify that

This change converts all values to strings for those modules and also adds a better error message for the HA module.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cluster_ha
cluster_drs

##### ADDITIONAL INFORMATION
Tests need https://github.com/ansible-collections/vmware.vmware/pull/191

### DIFF
--- a/plugins/modules/cluster_ha.py
+++ b/plugins/modules/cluster_ha.py
@@ -69,6 +69,8 @@ options:
     advanced_settings:
         description:
             - A dictionary of advanced HA settings.
+            - Allowed HA settings are more strict than VM allowed settings. If you get an error when managing them, compare
+              your settings against the vSphere HA Advanced Options documentation.
         default: {}
         type: dict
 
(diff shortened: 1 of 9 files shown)


p3ck (approved):
LGTM - But why is the eco vCenter integration test failing?  I looked at the error and it doesn't seem related to this.

plugins/modules/cluster_ha.py
```diff
@@ -684,7 +686,15 @@ def apply_ha_configuration(self):
         except (vmodl.RuntimeFault, vmodl.MethodFault)as vmodl_fault:
             self.module.fail_json(msg=to_native(vmodl_fault.msg))
         except TaskError as task_e:
-            self.module.fail_json(msg=to_native(task_e))
+            if not self.changed_advanced_settings.is_empty():
```
anna-savina, line 689: It might be worth adding this check to cluster_drs.py as well?

mikemorency: For some reason the DRS error message was informative, while the HA message was just empty. So I dont think we need to do anything special for DRS

anna-savina (approved):

PR #195, opened 2025-06-06
update inventory tests for 2.19

mikemorency:
##### SUMMARY
Fixes inventory tests for 2.19

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
inventory integration tests


mikemorency:
they are no longer actively working on this project. There is another developer coming up to speed, but I feel like you are close enough to this (and have been looking at it as long as I have) so your review is best.
Plus its just a test change :) we can always revert

PR #197, opened 2025-06-06
fix vmware_folder_template_from_vm tests

mikemorency:
##### SUMMARY
Fixes the tests for vmware_folder_template_from_vm and 2.19

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/integration/vmware_folder_template_from_vm

### DIFF
--- a/changelogs/fragments/197-fix-folder_template_from_vm-tests.yml
+++ b/changelogs/fragments/197-fix-folder_template_from_vm-tests.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - tests/integration/vmware_folder_template_from_vm - Fix tests for 2.19
--- a/tests/integration/requirements.txt
+++ b/tests/integration/requirements.txt
@@ -1,4 +1,4 @@
-ansible-core>=2.18.0,<2.19.0
+ansible-core==2.19.0b5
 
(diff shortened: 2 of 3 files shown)


mariolenz (commented):
Are you sure this would really fix any CI problem? You're just changing the name of two tasks, I don't see how this could fix anything.

What issue do you want to fix? I don't understand :shrug:

mikemorency:
The original failure said that `item` was not defined. I _believe_ the issue is that `item` is not an accessible variable when ansible parses the task title, but it is accessible by the time ansible parses the task parameters.

mariolenz (approved):
I think I've found this in the CI logs. I've only looked at errors but this is about a _warning_, right? I wonder if this change in behavior is really intended... maybe I'll ask around a bit.

Anyway: LGTM since it seems to get rid of the warning.

mariolenz:
> I wonder if this change in behavior is really intended... maybe I'll ask around a bit.

https://forum.ansible.com/t/40759/53

mikemorency:
That is a good point about it not making much sense, since the title is just printed once for the loop. I see that the "error" is actually just a warning, so maybe it has been complaining for a long time. 
https://github.com/ansible-collections/vmware.vmware/actions/runs/15499976187/job/43645500816?pr=188#step:5:1847

Issue #201, opened 2025-06-13
admission_control_failover_level cannot be set


pandalec:
##### SUMMARY

It is not possible to change the admission_control_failover_level. It is always == 1.

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
ansible [core 2.17.12]
```

##### COLLECTION VERSION

```paste below
vmware.vmware             2.1.0
```

##### CONFIGURATION

```paste below
ANY_ERRORS_FATAL(./ansible.cfg) = True
CALLBACKS_ENABLED(./ansible.cfg) = ['profile_tasks']
CONFIG_FILE() = ./ansible.cfg
DEFAULT_FORKS(./ansible.cfg) = 50
DEFAULT_JINJA2_NATIVE(./ansible.cfg) = True
DEFAULT_LOAD_CALLBACK_PLUGINS(./ansible.cfg) = True
DEFAULT_STDOUT_CALLBACK(./ansible.cfg) = debug
DISPLAY_SKIPPED_HOSTS(./ansible.cfg) = False
EDITOR(env: EDITOR) = hx
HOST_KEY_CHECKING(./ansible.cfg) = True
```

##### OS / ENVIRONMENT
Arch Linux

##### STEPS TO REPRODUCE

```yaml
    - name: Enable and configure HA
      vmware.vmware.cluster_ha:
        admission_control_failover_level: 2
```

Log shows:

```
"admission_control_failover_level": 2,
```

But `Host failures cluster tolerates:` in UI stays on `1`. If I change it manually to `2` in UI and rerun the play (with `admission_control_failover_level: 2`), it changes it back to `1`.

##### EXPECTED RESULTS

`Host failures cluster tolerates:` should have correct value

##### ACTUAL RESULTS

```paste below
`Host failures cluster tolerates:` is always == 1
```

mariolenz:
> It also sounds like the module is changing the failover level (from 2 to 1) when it shouldnt be changing anything. So that is something to verify and potentially fix.

I think `__set_admission_control_config` only uses this parameter when `admission_control_policy` is `vm_slots` or `cluster_resource`. So maybe the 1 comes from some default when creating the config that isn't overwritten?

mikemorency:
[...]

PR #203, opened 2025-06-14
add cluster admission policy required_by

mikemorency:
##### SUMMARY
Partial fix for https://github.com/ansible-collections/vmware.vmware/issues/201

The admission control policy is required if a user sets any other admission control options. This is mentioned in the docs but is otherwise un-enforced. The current behavior is to just ignore admission control options. This change adds `required_by` for admission control policy parameters so the module will alert the user instead

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cluster_ha


mikemorency:
Yea. I'm going to adjust and open a new one

PR #205, opened 2025-06-16
Add cluster_ha admission control required_by rules

mikemorency:
##### SUMMARY
Adds required_by rules for the cluster_ha module's admission control paramters. The documentation states that the policy must be provided for other options to take effect. This makes the module communicate that dependency to the user.

Fixes https://github.com/ansible-collections/vmware.vmware/issues/201

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cluster_ha


plugins/modules/cluster_ha.py
```diff
@@ -369,7 +369,7 @@ def ac_failover_level(self):
         Since this param requires admission_control_policy, we cannot have a default set in the module spec.
         But it does have a default value of 1, which is mentioned in the param docs
         """
-        self.params.get('admission_control_failover_level', 1)
+        return self.params.get('admission_control_failover_level', 1)
```
p3ck, line 372: Looks better. :-)

But I'm still confused why this is needed since it doesn't do any logic.   Is this going to be expanded in the future?

plugins/modules/cluster_ha.py
```diff
@@ -362,6 +363,14 @@ def ac_cluster_resource_auto_compute_percentages(self):
...
         return True
 
+    @property
+    def ac_failover_level(self):
+        """
+        Since this param requires admission_control_policy, we cannot have a default set in the module spec.
+        But it does have a default value of 1, which is mentioned in the param docs
+        """
+        self.params.get('admission_control_failover_level', 1)
+
```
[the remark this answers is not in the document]
mikemorency: Ah wow I am glad you caught that, I thought I was going insane when the tests failed

plugins/modules/cluster_ha.py
mikemorency: No, it wont be exanded. We had 1 as the default value in the module spec, but that caused the new `required_by` rules to trigger every time. So I thought making it a property would be best, instead of having `self.params.get('admission_control_failover_level', 1)` in the two or three places that the parameter is used. WDYT?

p3ck: ahhh ok..  Yes - That makes sense.  I was reading your comment too!  It just didn't click.  Or I guess I didn't realize the default would cause ansible to behave that way.  But in hindsight I can't see how it wouldn't. :-)

All good.  Carry on..

p3ck (approved):

PR #209, opened 2025-06-18
allow failover level to always be managed

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/201

This change allows the user to set the failover host threshold setting, no matter what admission control policy they use. This is how it works in the vSphere UI.

Previously, when using a dedicated failover host policy (the user lists the hosts that are on reserve for when another host fails) the failover threshold would always be set to 1. This stopped users from controlling the value and doesnt make sense as a default. The default should be the number of hosts that are dedicated for failover, since that is the number of hosts you have ready to replace failed hosts.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
cluster_ha

##### ADDITIONAL INFORMATION
I was not able to find any documentation confirming if the failover host threshold is even considered by vSphere with dedicated host policy mode or not. Documentation in general is lacking for HA admission control


mikemorency:
opened https://github.com/ansible-collections/vmware.vmware/pull/210 for the test failures

Issue #207, opened 2025-06-17
disable proxy doesn't work


pugnacity:
##### SUMMARY
removing a proxy setting doesn't work

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
vmware.vmware.vcsa_setting

##### ANSIBLE VERSION

```
ansible [core 2.18.6]
  config file = /home/pugnacity/repos/project/project-ansible/ansible.cfg
  configured module search path = ['/home/pugnacity/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules', '/home/pugnacity/repos/project/project-ansible/modules']
  ansible python module location = /home/pugnacity/bin/ansible/lib64/python3.11/site-packages/ansible
  ansible collection location = /home/pugnacity/repos/project/project-ansible/collections:/home/pugnacity/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/pugnacity/bin/ansible/bin/ansible
  python version = 3.11.2 (main, Sep 17 2024, 03:17:19) [GCC 8.5.0 20210514 (Red Hat 8.5.0-18)] (/home/pugnacity/bin/ansible/bin/python3.11)
  jinja version = 3.1.6
  libyaml = True

```

##### COLLECTION VERSION

```paste below
# /home/pugnacity/bin/ansible/lib/python3.11/site-packages/ansible_collections
Collection    Version
------------- -------
vmware.vmware 1.11.0 

# /home/pugnacity/bin/ansible/lib64/python3.11/site-packages/ansible_collections
Collection    Version
------------- -------
vmware.vmware 1.11.0 

# /home/pugnacity/repos/project/project-ansible/collections/ansible_collections
Collection    Version
------------- -------
vmware.vmware 2.0.1  
```

##### CONFIGURATION

```paste below
CACHE_PLUGIN(/home/pugnacity/repos/project/project-ansible/ansible.cfg) = ansible.builtin.jsonfile
CACHE_PLUGIN_CONNECTION(/home/pugnacity/repos/project/project-ansible/ansible.cfg) = ~/.cache/facts_cache
CACHE_PLUGIN_TIMEOUT(/home/pugnacity/repos/project/project-ansible/ansible.cfg) = 300
CALLBACKS_ENABLED(/home/pugnacity/repos/project/project-ansible/ansible.cfg) = ['ansible.posix.profile_tasks', 'ansible.posix.timer']
[...]

PR #219, opened 2025-07-10
fix issue where vcsa proxies could not be disabled

mikemorency:
##### SUMMARY
Fixes https://github.com/ansible-collections/vmware.vmware/issues/207

When a user sets the VCSA proxy enabled paramter to false, the module skips the proxy config. This means users cannot disable proxies once enabled.

This change allows users to disable the proxies. I also added some tests around the proxy functionality, since its relatively safe to test in our vcenter lab

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vcsa_settings


plugins/modules/vcsa_settings.py
```diff
@@ -424,20 +424,27 @@ def __proxy(self):
...
+            if p['enabled']:
+                change_required = any([
+                    c.enabled != p['enabled'],
+                    c.server != p['url'],
+                    c.port != p['port'],
+                    getattr(c, 'username', None) != p.get('username', None),
+                    p.get('password', None) is not None
+                ])
+            else:
+                change_required = bool(c.enabled != p['enabled'])
```
mariolenz, line 437:
```suggestion
            change_required = any([
                bool(c.enabled != p['enabled']),
                c.enabled != p['enabled'],
                c.server != p['url'],
                c.port != p['port'],
                getattr(c, 'username', None) != p.get('username', None),
                p.get('password', None) is not None
            ])
```

I don't understand why you're doing a `if p['enabled']` / `else` here. Wouldn't it be simpler to do this all in one go? Or am I overlooking something? Is there some corner case that this approach wouldn't address?

And why is it `p.get('password', None) is not None` instead of `p.get('password', None) is not None` and not `getattr(c, 'password', None) != p.get('password', None),`

@p3ck What do you think? A second opinion on this might be helpful.

mariolenz: Thanks for the explanation, makes sense to me.

> The API does say it returns a password, but I know for other endpoints the password value is encrypted or not returned at all. I suppose I should investigate a bit more to confirm if we can compare the two values or not

I could also investigate this. But give me a few days. I'm pretty sure I won't be able to this tomorrow, but maybe the day after.

If the API returns the password, I think it would be better to compare it with the one given in the task. Otherwise, we'll always get a change when when a password is given. That's not really idempotent. If the API doesn't return the password, we can't do anything about it. But if it does, we can improve idempotency there :-)

mikemorency: I am in no rush and appreciate the help :)

mariolenz: I'm afraid we won't be able to check the current password against the one defined in the task. I've tested this today, and the API doesn't return the current password.

I think they're just re-using the data structure for _setting_ the configuration. It contains `password`, but this is always empty / `None`. But they don't return the password, probably for security reasons. At least this is the result from my tests.

mikemorency: OK. Seems consistent with other API endpoints at least. I will update the module docs to reflect this

mikemorency: Added an option that other modules have which allow sthe user to control how the password is handled. The default behavior is consistent with the current behavior. The module will always try to update the password since it cant check the value

mariolenz (approved):
LGTM

PR #229, opened 2025-08-26
Feature/vm track changes

mikemorency:
##### SUMMARY
This adds more details to the vm module output when parameters are changed. When something has changed the output includes the relevant module parameter, the old value, and the new value. It is only relevant when an existing vm is being updated.

I am adding this now because not knowing what parameter is causing the module to report a change makes debugging difficult. I think it is also a useful feature to have for the module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
_vm_stub


p3ck (approved):
lgtm

PR #230, opened 2025-08-31
Drop aiohttp requirement

mariolenz:
##### SUMMARY
It looks like the collection defines a requirement on aiohttp, but it doesn't use it anywhere:

```
$ grep -r aiohttp plugins/
plugins/modules/deploy_content_library_template.py:    - aiohttp
plugins/modules/deploy_content_library_ovf.py:    - aiohttp
$
```

You see, this package is documented as a requirement but not used anywhere.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
deploy_content_library_ovf
deploy_content_library_template
requirements.txt
tests/integration/requirements.txt
tests/unit/requirements.txt

##### ADDITIONAL INFORMATION
https://forum.ansible.com/t/44392

### DIFF
--- a/changelogs/fragments/230-drop-aiohttp-requirement.yml
+++ b/changelogs/fragments/230-drop-aiohttp-requirement.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - Drop incorrect requirement on aiohttp (https://github.com/ansible-collections/vmware.vmware/pull/230).
--- a/plugins/modules/deploy_content_library_ovf.py
+++ b/plugins/modules/deploy_content_library_ovf.py
@@ -22,7 +22,6 @@ author:
 
 requirements:
     - vSphere Automation SDK
-    - aiohttp
 
 extends_documentation_fragment:
     - vmware.vmware.base_options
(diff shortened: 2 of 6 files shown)


mariolenz:
@mikemorency Do you have an idea why the sanity tests fail? It doesn't look directly related to this PR to me, but because of some changes in the ansible-core devel branch.

mariolenz:
What should we do about this? Drop sanity tests with devel for now, or at least make them non-voting? Or do you have an idea how to fix this by not using `six`? Because ATM I don't :-/

mikemorency:
It sounds like the packages we were importing from six can be imported directly from the standard python 3 libs. So should just be an import statement update

I can open a pr later today or tomorrow if you are ok waiting. Otherwise you can add an [ignore.txt](https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/ignores.html) file for the devel sanity tests and I'll remove it in my pr

mariolenz:
I don't think we'll need a quick'n'dirty workaround like ignoring those issues for this PR. It can wait until the sanity issues with ansible-core devel have been fixed.

At the end of the day, I think this PR is desirable but not _really_ important. It can wait.

mikemorency (approved):

PR #231, opened 2025-09-01
make all parameters optional in module spec

mikemorency:
##### SUMMARY
This change makes all vm config parameters optional in the module spec. 

No VM config parameters should be required to update the VM. Parameter handlers can check if a relevant parameter name ( for example, the CpuHandler can check the parameter 'cpu') is defined, and throw an error if the parameter is not defined but required for VM creation.

Parameter handlers now have a flag that indicates if their relevant parameter was defined by the user. If a parameter handler's parameter was not defined, it will not be included in the configuration builder. This will hopefully prevent unnecessary code execution

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
vm_stub


p3ck (approved):
LGTM

PR #240, opened 2025-09-15
update python reqs

mikemorency:
##### SUMMARY
Update python requirements to match what the collection integration tests use. People have expressed that using a git requirement is not ideal, but that is the only version of that dependency that we test and know works.

### DIFF
--- a/changelogs/fragments/239-fix-collections-py-reqs.yml
+++ b/changelogs/fragments/239-fix-collections-py-reqs.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Update collection python requirements to match what integration tests are using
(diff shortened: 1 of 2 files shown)


mariolenz:
We use the following modules in our playbook:

- cluster
- cluster_drs
- cluster_info
- vcsa_settings
- vmware_dvswitch_info
- vmware_host_lockdown_exceptions

They seem to work fine with the following packages installed:

- pyVmomi>=8.0.3.0.1
- vmware-vapi-common-client
- vmware-vcenter

Of course, there might be a problem in some part of the code that wasn't executed when I ran my test. Or in some other module.

I've tested with the version that's included in Ansible 12, that is 2.3.0.

mariolenz:
It looks like the integration tests are quite happy without `git+https://github.com/vmware/vsphere-automation-sdk-python.git@v8.0.2.0`: #241

Maybe there have been problems in the past, but before we introduce a dependency on a GitHub repo again I think we should make sure that they still exist.

PR #241, opened 2025-09-16
Update requirements

mariolenz:
##### SUMMARY
Update integration test requirements.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
requirements.txt
tests/integration/requirements.txt
tests/unit/requirements.txt

##### ADDITIONAL INFORMATION

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
 pyVmomi>=8.0.3.0.1
 vmware-vcenter
-vmware-vapi-common-client
--- a/tests/integration/requirements.txt
+++ b/tests/integration/requirements.txt
@@ -1,5 +1,6 @@
 ansible-core
 
+aiohttp
 pyVmomi>=8.0.3.0.1
 vmware-vcenter
 vmware-vapi-common-client
--- a/tests/integration/requirements.txt
+++ b/tests/integration/requirements.txt
@@ -1,6 +1,5 @@
-ansible-core==2.19.0b5
+ansible-core
 
-git+https://github.com/vmware/vsphere-automation-sdk-python.git@v8.0.2.0
-podman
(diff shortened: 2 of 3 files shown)


mikemorency:
I am pretty sure this test is not valid, because we need to set the ANSIBLE_REMOTE_INTERPRETER var somewhere in the CI (i think the gh workflow makes the most sense but then we need to merge that change first).

I know for a fact the integration tests need aiohttp (for vmware.vmware_rest modules). So if the tests pass, then this current test is using the wrong python env

mariolenz:
> I am pretty sure this test is not valid, because we need to set the ANSIBLE_REMOTE_INTERPRETER var somewhere in the CI (i think the gh workflow makes the most sense but then we need to merge that change first).

Is ANSIBLE_REMOTE_INTERPRETER something outside of this repo? I didn't find it anywhere in the code.

> I know for a fact the integration tests need aiohttp (for vmware.vmware_rest modules). So if the tests pass, then this current test is using the wrong python env

That's correct, the integration tests failed when I removed it from `tests/integration/requirements.txt`. When I've added `aiohttp` again, they succeeded.

The good thing is: They still succeed with just `ansible-core` instead of `ansible-core==2.19.0b5`, `pyVmomi>=8.0.3.0.1` + `vmware-vcenter` (_without_ `vmware-vapi-common-client`) instead of `git+https://github.com/vmware/vsphere-automation-sdk-python.git@v8.0.2.0`, and dropping `podman` and `requests`.

So I think we can safely simplify the requirements :smiley:

mikemorency (approved):

mariolenz:
Thanks for reviewing @mikemorency!

Issue #243, opened 2025-09-19
adminission_control_policy ignored  in some cases


henry-nicolastourneur-blip:
##### SUMMARY
Under some condition, the adminission_control_policy is ignored.
While the vCenter UI exposes 4 options (3 policies and disabled), when switching a cluster to disabled, the vCenter retains the policy that was set before switching to disabled.
In that sense the UI is misleading because the cluster has in fact 2 settings combined in a single drop-down:
- whether or not admissionControl is enabled (dasConfig.admissionControlEnabled)
- one of the 3 policy when enabled (vm_slots, cluster_resource, dedicated_host)  

That is counter-intuitive: one would expect that setting a policy using ansible also enables admission_control.
Instead, the ansible module only checks for the policy match and if the vCenter retained policy is matched (before switching to disable), nothing is changed.
The steps to reproduce might help better understand the issue.
I'll open a PR shortly with a fix for this issue.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cluster_ha / admission_control_policy
##### ANSIBLE VERSION
```
ansible [core 2.11.12]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/htourneu/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/site-packages/ansible
  ansible collection location = /home/htourneu/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.6.8 (default, Sep  9 2024, 06:12:39) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)]
  jinja version = 2.10.1
  libyaml = True

```

##### COLLECTION VERSION
```
# /usr/share/ansible/collections/ansible_collections

Collection    Version
------------- -------
vmware.vmware 2.1.0
```

##### CONFIGURATION
```
... empty output here

```

##### OS / ENVIRONMENT
RHEL 8.10

##### STEPS TO REPRODUCE
1. Run the playbook task below to enable admission control with vm_slots policy.
2. Login in vCenter UI and update the cluster to disable admission control.
3. Re-run the playbook from step 1
4. Observe that ansible doesn't change anything on the cluster.
    Note: this occurs because we only compare the 3 policy classes options and not the global status "admissionControlEnabled"

```yaml
   - name: Set HA settings on cluster {{ item.key }}
     vmware.vmware.cluster_ha:
[...]

PR #244, opened 2025-09-19
fix: cluster admission control: set requested policy also when ac is disabled

henry-nicolastourneur-blip:
If the cluster admission control (ac.) policy is set and then ac. is disabled, we want to reinstate the policy setting and enable ac.
See bug for details on how to reproduce.

Fixes: #243


mikemorency:
Please add a changelog file. You can find examples in `changelogs/fragments/`

PR #246, opened 2025-09-23
Add module for uploading ISO images to content library.

miikkajo:
##### SUMMARY
Current modules only allow uploading of OVF or OVA images to content library.

This module allows user to upload or import ISO images to content library.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

import_content_library_iso

##### ADDITIONAL INFORMATION

Currently user can only import OVF or OVA images to content library using import_content_library_ovf,
I am using content library also for installer iso images and kickstart images.

This feature adds new module import_content_library_iso to allow uploading or importing of ISO images.

As this module is pretty much copy of import_content_library_ovf module,
it's debatable, should this be separate module or should import_content_library_ovf to be improved.
A new module with a lesser impact on old automations seems like a better option to me.

### DIFF
--- a/changelogs/fragments/246-add-import-content-library-iso.yml
+++ b/changelogs/fragments/246-add-import-content-library-iso.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Add module for importing iso images to content library.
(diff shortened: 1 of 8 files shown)


mikemorency:
Please fix the test failures and add a changelog file (see changelog/fragments/ for examples. This would go under `minor_changes`). It would be great if you could add unit tests too. Seems like it is mostly just copying the existing tests. Otherwise, I will add them to the PR when I have some time
Once all that is done we can merge

PR #248, opened 2025-09-30
Feature/vm cdrom

mikemorency:
##### SUMMARY
Adding CDROM params to the vm module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
vm_stub

### DIFF
--- a/tests/integration/targets/vmware_vm_stub/tasks/main.yml
+++ b/tests/integration/targets/vmware_vm_stub/tasks/main.yml
@@ -45,12 +45,6 @@
     - name: Test network adapter params
       ansible.builtin.include_tasks: network_adapter_params_tests.yml
 
-    - name: Test memory params
-      ansible.builtin.include_tasks: mem_params_tests.yml
-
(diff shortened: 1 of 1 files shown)


tests/integration/targets/vmware_vm_stub/defaults/main.yml
```diff
@@ -1,2 +1,4 @@
 ---
+#test_vm_name: "{{ tiny_prefix }}-vm-stub"
```
p3ck, line 2: Did you forget to make this active or do you really want test_vm_stub as the test_vm_name?

p3ck (approved):

tests/integration/targets/vmware_vm_stub/defaults/main.yml
mikemorency: Ah I did not mean to leave it commented out. Shouldn't cause problems though, so I'll fix it in the next PR

PR #254, opened 2025-10-03
update category outputs to include ids

mikemorency:
##### SUMMARY
Changing the output of the tag module to better work with other modules, like tag_associations.
These modules benefit from using IDs instead of names, so it makes sense to provide the user with the tags identifiers as output when possible.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
tags

### DIFF
--- a/plugins/modules/tag_categories.py
+++ b/plugins/modules/tag_categories.py
@@ -157,30 +157,51 @@ EXAMPLES = r"""
 """
 
 RETURN = r"""
-category_changes:
+vsphere_tag_categories:
     description:
-        - Comparison of the tag categories before and after the changes.
-        - Before is empty if the tag category was created.
-        - After is empty if the tag category was deleted.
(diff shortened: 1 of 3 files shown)


plugins/modules/tag_categories.py
```diff
@@ -527,20 +557,37 @@ def main():
...
+        removed_categories=[],
+        vsphere_tag_categories={},
+    )
 
     vmware_tag_category = VmwareTagCategoryModule(module)
-    tag_category_changes = vmware_tag_category.determine_tag_category_changes()
-    if not tag_category_changes:
+    tag_category_diffs = vmware_tag_category.determine_tag_category_diffs()
+    result["vsphere_tag_categories"] = {
+        cat_diff.get_category_id(): cat_diff.new_tag_category_state_to_module_output()
```
p3ck, line 571: Am I missing something?  I don't see cat_diff defined yet.

PR #295, opened 2025-12-16
automatically use system certs via truststore

omenos:
##### SUMMARY

`requests` defaults to using the certificate trust provided by `certifi`, which precludes the inclusion of custom trusted certificates. Rather than needing to set the `REQUESTS_CA_BUNDLE` environment variable in every shell/config, use [`truststore`](https://github.com/sethmlarson/truststore) to override the way certs are read and automatically use the system trust store.

This will only happen transparently in environments using [Python >= 3.10](https://github.com/sethmlarson/truststore/blob/main/src/truststore/__init__.py).

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/plugins/module_utils/clients/pyvmomi.py
+++ b/plugins/module_utils/clients/pyvmomi.py
@@ -11,6 +11,12 @@ import atexit
 import traceback
 
 try:
+    import truststore
+    truststore.inject_into_ssl()
+except ImportError:
+    pass
+
+try:
     # requests is required for exception handling of the ConnectionError
     import requests
     REQUESTS_IMP_ERR = None
--- a/plugins/module_utils/clients/rest.py
+++ b/plugins/module_utils/clients/rest.py
@@ -14,6 +14,12 @@ __metaclass__ = type
 import traceback
 
 try:
+    import truststore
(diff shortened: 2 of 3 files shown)


mikemorency (changes requested):
This change would force a user to use the system bundle, right? They have no way of choosing a third, custom CA bundle (if they are not supposed to use system bundles or third party bundles, for example)

Additionally, it changes the expected behavior for existing users. Instead of using the certifi store, they will suddenly be using the system store. We would need a major version change to include this

Unfortunately I dont think we will be able to include this change as is

omenos:
I just did a quick check (on RHEL) and if `SSL_CERT_FILE` is specified, it will be used rather than the system store.

So it's a bit of a behavior reversal, rather than:

* Default: `certifi`
* `REQUESTS_CA_BUNDLE`: selected bundle

it becomes:

* Default: system store via `truststore`
* `SSL_CERT_FILE`: selected bundle

I understand this likely requiring a major jump due to the difference, this is just a submission to kind of stir the pot on the idea.

mikemorency:
Im going to close this since we are not planning on implementing it and the branch has gone stale

PR #324, opened 2026-03-09
Replace wildcard catch-all with per-module queries in event_query.yml

jonnyfiveiq:
**Title:** Replace wildcard catch-all with per-module queries in event_query.yml

**Description:**

## Summary

The current `event_query.yml` uses a `vmware.vmware.*` wildcard catch-all alongside a specific `vmware.vmware.guest_info` entry. This causes a silent failure in AAP 2.5 indirect managed node counting where `bulk_create()` rolls back the entire batch to zero records when both queries match events for the same VM.

This PR replaces the 2 entries (specific + wildcard) with 38 per-module query entries, eliminating the conflict and expanding coverage to all modules in the collection.

## Problem

When a playbook uses `guest_info` alongside any other vmware.vmware module (e.g. `vm_powerstate`) on the same VM, the specific `guest_info` query produces a record with `canonical_facts: {instance_uuid, bios_uuid, moid, esxi}` while the `*` catch-all produces a record with `canonical_facts: {name, id, node_type}`. The deduplication in `build_indirect_host_data()` treats these as separate records because the `hashable_facts` differ. However both records share the same `name` value (the VM moid), so `bulk_create()` fails on the `(name, job_id)` unique constraint in `main_indirectmanagednodeaudit` and silently rolls back the entire batch — saving zero indirect nodes for that job.

The `event_queries_processed` flag is then set to `True`, so the job is never reprocessed.

Analysis of 89 SaaS customer clusters confirmed widespread zero indirect node counts despite active VMware automation.

## Changes

- Removed the `vmware.vmware.*` wildcard catch-all entry
- Removed the existing `vmware.vmware.guest_info` entry with its overly complex `canonical_facts`
- Added 38 per-module jq query entries covering all vmware.vmware modules (guest_info, vm_powerstate, vm, cluster, esxi_host, folder, etc.)
- Each entry uses consistent `canonical_facts` including a `module` field to ensure distinct records per module action
- Module name is included in the `name` field (e.g. `vm_powerstate:vm-15656`) to avoid the `(name, job_id)` unique constraint collision
- Each query is tailored to its module's actual return data structure (`.vm`, `.guests[]`, `.cluster`, `.host`, etc.) rather than relying on a fragile generic catch-all

## Root Cause Detail

The catch-all query assumed all modules return data under a top-level key matching one of `vm`, `host`, `vcsa`, `cluster` with `.name` and `.moid` sub-fields. This was fragile and also structurally incompatible with the specific `guest_info` query, creating the duplicate name / different canonical_facts combination that triggers the `bulk_create()` rollback.

## Testing

Tested in a lab environment running AAP controller 4.6.15 with a playbook that creates a VM (`community.vmware.vmware_guest`), powers it off (`vmware.vmware.vm_powerstate`), gathers info (`vmware.vmware.guest_info`), and deletes it. With the upstream wildcard query, zero indirect node records were saved. With the fixed per-module queries, all three non-delete tasks correctly produced distinct records in `main_indirectmanagednodeaudit`.

### DIFF
--- a/extensions/audit/event_query.yml
+++ b/extensions/audit/event_query.yml
@@ -1,13 +1,249 @@
 ---
-vmware.vmware.guest_info:
+vmware.vmware.appliance_info:
   query: >-
-    .guests[] | {
-      name: .moid,
+    .appliance | select(. != null) | {
+      name: ("appliance_info:" + (.id| tostring)),
       canonical_facts: {
-        instance_uuid: .instance_uuid,
-        bios_uuid: .hw_product_uuid,
-        moid: .moid,
-        esxi: .hw_esxi_host
+            module: "appliance_info",
+        id: .id
+      },
+      facts: {
+        infra_type: "PrivateCloud",
+        infra_bucket: "Compute",
+        device_type: "vCenter Appliance"
+      }
+    }
+vmware.vmware.cluster:
+  query: >-
+    .cluster | select(. != null) | {
+      name: ("cluster:" + (.moid| tostring)),
+      canonical_facts: {
+            module: "cluster",
+        moid: .moid
+      },
+      facts: {
+        infra_type: "PrivateCloud",
+        infra_bucket: "Compute",
+        device_type: "Cluster"
+      }
+    }
+vmware.vmware.cluster_dpm:
+  query: >-
+    .cluster | select(. != null) | {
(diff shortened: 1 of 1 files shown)


p3ck (approved):
This looks good to me.  I'm trying to resolve some CI issues that have risen because upstream hasn't updated the python library to work without pkg_resources.

See #325

Issue #330, opened 2026-03-25
vm_snapshot_revert: Can't revert to snapshots in non-trivial locations


michal-zatloukal-eset:
##### SUMMARY
vm_snapshot_revert doesn't properly search the snapshot tree

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
vm_snapshot_revert

##### ANSIBLE VERSION

```
ansible [core 2.20.3]
  config file = /home/jenkins/git/ansible/ansible.cfg
  configured module search path = ['/home/jenkins/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.12/site-packages/ansible
  ansible collection location = /home/jenkins/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.12.7 (main, Jan 24 2025, 13:07:00) [GCC 12.2.0] (/usr/local/bin/python3.12)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION

```
# /usr/local/lib/python3.12/site-packages/ansible_collections
Collection    Version
------------- -------
vmware.vmware 2.7.0
```

##### CONFIGURATION

```
CONFIG_FILE() = /home/jenkins/git/ansible/ansible.cfg
DEFAULT_DEBUG(env: ANSIBLE_DEBUG) = True
DEFAULT_KEEP_REMOTE_FILES(env: ANSIBLE_KEEP_REMOTE_FILES) = True
DEFAULT_STDOUT_CALLBACK(env: ANSIBLE_STDOUT_CALLBACK) = debug
DISPLAY_SKIPPED_HOSTS(/home/jenkins/git/ansible/ansible.cfg) = False
HOST_KEY_CHECKING(/home/jenkins/git/ansible/ansible.cfg) = False
INJECT_FACTS_AS_VARS(/home/jenkins/git/ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/home/jenkins/git/ansible/ansible.cfg) = auto
GALAXY_SERVERS:
```

##### OS / ENVIRONMENT

container based on Debian 12

##### STEPS TO REPRODUCE

- prepare a non-trivial snapshot tree on the target VM:
```
snapshots
└── snap1
       ├── snap1A
      └── snap1B
```
Revert to snapshot snap1B

```yaml
---
[...]

PR #331, opened 2026-03-26
Fix issue #330, traverse snapshot tree correctly in vm_snapshot_revert

michal-zatloukal-eset:
vm_snaphot_revert/_get_snapshot_by_identifier_recursively: Don't return  None until all snapshots from list argument are exhausted.

##### SUMMARY

Fix issue #330 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

vm_snapshot_revert

##### ADDITIONAL INFORMATION

The previous form of _get_snapshot_by_identifier_recursively never iterates over the list that is passed into the function, because both branches of the `if..else` construct issue a `return`.

Fix tested with ansible 2.20, applied to vmware collection 2.7.0

```paste below

```

### DIFF
--- a/changelogs/fragments/bugfix-330.yml
+++ b/changelogs/fragments/bugfix-330.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - vm_snapshot_revert - search the snapshot tree fully to find a matching snapshot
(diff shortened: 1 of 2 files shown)


p3ck:
Can you update the PR to include a changelog fragment?  Thanks

p3ck (approved):

PR #340, opened 2026-04-28
feat(license_info): export detailed license information

brosenqvist:
##### SUMMARY
Enhances vmware.vmware.license_info to return detailed, structured license data instead of only license keys.

This change adds:
- A richer licenses schema with fields such as name, license_key, edition_key, cost_unit, total, used, expiration_date, and labels
- A new include_properties module option (default: false) to optionally expose normalized vCenter license properties
- Normalization helpers to safely convert pyVmomi values into stable JSON-serializable output and avoid overly verbose opaque objects
- Updated module docs, examples, and RETURN contract to reflect the new behavior

Rationale:
- Existing output was too minimal for operational and reporting use cases
- Consumers need consistent fields for automation and inventory/report pipelines
- Optional property expansion keeps default output concise while allowing deeper inspection when needed

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
license_info

##### ADDITIONAL INFORMATION
Behavioral notes:
- Module output is now a list of structured objects rather than a flat list of license keys
- include_properties=true includes normalized vCenter property data per license
- Expiration is inferred from common property keys when present (expirationDate, expiration_date, expiryDate, expiry_date)
- Empty/null values are filtered from normalized property payloads to keep output clean

Changed file:
- plugins/modules/license_info.py

### DIFF
--- a/plugins/modules/license_info.py
+++ b/plugins/modules/license_info.py
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+# -*- coding: utf-8 -*-
 
 # Copyright: (c) 2024, Ansible Cloud Team (@ansible-collections)
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
@@ -17,6 +18,13 @@ author:
 - Ansible Cloud Team (@ansible-collections)
 requirements:
 - Python SDK for the VMware vSphere Management API
+options:
+    include_properties:
+        description:
+        - Include normalized license properties in output.
+        - When true, returns all normalized property items.
+        type: bool
+        default: false
 attributes:
   check_mode:
     description: The check_mode support.
@@ -26,11 +34,18 @@ extends_documentation_fragment:
 '''
 
 EXAMPLES = r'''
-- name: Fetch vCenter license
-  vmware.vmware.license_info:
-    hostname: '{{ vcenter_hostname }}'
-    username: '{{ vcenter_username }}'
-    password: '{{ vcenter_password }}'
+- name: Fetch vCenter license details
+    vmware.vmware.license_info:
+        hostname: '{{ vcenter_hostname }}'
+        username: '{{ vcenter_username }}'
+        password: '{{ vcenter_password }}'
+
+- name: Fetch vCenter license details with properties
+    vmware.vmware.license_info:
+        hostname: '{{ vcenter_hostname }}'
+        username: '{{ vcenter_username }}'
+        password: '{{ vcenter_password }}'
+        include_properties: true
 '''
 
 RETURN = r'''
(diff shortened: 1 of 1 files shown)


plugins/modules/license_info.py
```diff
@@ -26,62 +34,237 @@
 '''
 
 EXAMPLES = r'''
-- name: Fetch vCenter license
-  vmware.vmware.license_info:
-    hostname: '{{ vcenter_hostname }}'
-    username: '{{ vcenter_username }}'
-    password: '{{ vcenter_password }}'
+- name: Fetch vCenter license details
+    vmware.vmware.license_info:
```
mikemorency, line 38: the indentation in the examples seems to be off

plugins/modules/license_info.py
```diff
@@ -26,62 +34,237 @@
...
+        hostname: '{{ vcenter_hostname }}'
+        username: '{{ vcenter_username }}'
+        password: '{{ vcenter_password }}'
+        include_properties: true
 '''
 
 RETURN = r'''
 licenses:
     description: List of license keys.
```
mikemorency, line 53: I think the returned value description should be updated to indicate that if you do not use the properties parameter, the returned value will be a list of strings (license keys).

plugins/modules/license_info.py
```diff
@@ -26,62 +34,237 @@
...
 from ansible_collections.vmware.vmware.plugins.module_utils.argument_spec import (
-    base_argument_spec
+    base_argument_spec,
 )
 
 
 class VcenterLicenseMgr(ModulePyvmomiBase):
     def __init__(self, module):
         super(VcenterLicenseMgr, self).__init__(module)
+        self.include_properties = module.params.get("include_properties")
```
mikemorency, line 121: if your only using this parameter reference once, I would suggest to not store it as a class attribute

plugins/modules/license_info.py
```diff
@@ -26,62 +34,237 @@
...
+            labels = self._labels_to_dict(getattr(item, "labels", None))
+            entry = {
+                "name": self._normalize_value(getattr(item, "name", None)),
+                "license_key": self._normalize_value(getattr(item, "licenseKey", None)),
+                "edition_key": self._normalize_value(getattr(item, "editionKey", None)),
+                "cost_unit": self._normalize_value(getattr(item, "costUnit", None)),
+                "total": self._normalize_value(getattr(item, "total", None)),
+                "used": self._normalize_value(getattr(item, "used", None)),
+                "expiration_date": self._guess_expiration(properties),
+                "labels": labels,
```
mikemorency, line 235: It seems like there is a lot of extra formatting/normalizing that is not needed since Broadcom provides the types for these attributes. For example, costUnit is always provided as a string. 
The one exception is `properties`, and it seems to need a lot of additional code/complexity. You might want to look at the util `vmware_obj_to_json` or other methods in `_facts.py` and see if that helps serialize the properties well enough.

mikemorency (changes requested):
You cant change the inputs or outputs of a module without a breaking change/major version bump. So you should maintain the current functionality of the module, and use an optional parameter to switch output from a list of keys to a more complex dict (or always return the list of keys and optionally include the enriched output in a different return value)

PR #348, opened 2026-05-07
fix flakey tests

mikemorency:
##### SUMMARY
Fixes for the more flakey tests:
- Adding steps to wait for hosts to be ready in the customization tests to ensure ssh is running
- Changing how we get the timestamp for the powerstate test to `now` instead of relying on the timestamp from when facts were gathered

##### ISSUE TYPE
- Bugfix Pull Request


p3ck (approved):
nice

PR #352, opened 2026-05-12
feat: add cluster_ha_vm_overrides modules and shared HA settings utils

mikemorency:
#### SUMMARY

This change adds `cluster_ha_vm_overrides` and `cluster_ha_vm_overrides_info` so clusters can manage and report HA VM override settings. 

Additionally, 
- shared logic between these modules, `cluster_ha` and the DRS override modules was moved into `plugins/module_utils/_cluster_settings.py` to avoid duplication.
- `cluster_drs_vm_overrides` and `cluster_drs_vm_overrides_info` were refactored to align with the same patterns as the new HA override modules. 

#### ISSUE TYPE

- Feature Pull Request

#### COMPONENT NAME

- `plugins/module_utils/_cluster_settings.py`
- `plugins/modules/cluster_ha.py`
- `plugins/modules/cluster_ha_vm_overrides.py`
- `plugins/modules/cluster_ha_vm_overrides_info.py`
- `plugins/modules/cluster_drs_vm_overrides.py`
- `plugins/modules/cluster_drs_vm_overrides_info.py`

#### ADDITIONAL INFORMATION

- Related internal ticket context: ACA-2283.

*This PR description was drafted with AI assistance and reviewed by the author.*

### DIFF
--- a/changelogs/fragments/aca-2283-cluster-ha-vm-overrides.yml
+++ b/changelogs/fragments/aca-2283-cluster-ha-vm-overrides.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - cluster_ha_vm_overrides - add module to manage HA VM override settings on vSphere clusters
+  - cluster_ha_vm_overrides_info - add module to gather HA VM override settings for a cluster
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -32,3 +32,4 @@ build_ignore:
   - .gemini/
   - .claude/
   - .ansible-lint
+  - .ansible/
--- a/plugins/module_utils/_cluster_settings.py
+++ b/plugins/module_utils/_cluster_settings.py
@@ -37,10 +37,16 @@ class ClusterSettingsRemapper:
             param_value: Truthy to restart VMs after APD clears; falsy to take no reaction.
 
         Returns:
-            ``reset`` if ``param_value`` is true, else ``none``.
+            ``reset`` if ``param_value`` is true, ``none`` if false, or ``None`` if unset
+            (caller should omit the API field when ``None``).
         """
+        # return None if the param was not set, to preserve user input
+        if param_value is None:
+            return None
+
         if param_value:
             return "reset"
+
         return "none"
 
 
@@ -80,6 +86,7 @@ class BaseVmOverrideChangeTracker(ABC):
         """
         if getattr(desired_spec, attribute_name) is None:
             return False
+
         return getattr(desired_spec, attribute_name) != getattr(
             current_spec, attribute_name
         )
--- a/plugins/module_utils/_cluster_settings.py
+++ b/plugins/module_utils/_cluster_settings.py
@@ -1,0 +1,153 @@
+"""
+Helpers for cluster HA and HA/DRS VM override modules: value remapping for the vSphere API
+and a shared change tracker for desired vs current per-VM override specs.
+"""
+
+from abc import ABC, abstractmethod
+
+
+class ClusterSettingsRemapper:
+    """
+    Remap cluster settings from the common module parameter values to the expected
+    vSphere API values.
+    This is done to provide a consistent and intuitive interface for the module(s).
+    """
+
+    @staticmethod
+    def storage_pdl_response_mode(param_value):
+        """
+        Map module PDL storage protection parameter values to vim API enum strings.
+
+        Args:
+            param_value: User-facing value such as ``restart``, ``warning``, or ``disabled``.
+
+        Returns:
+            ``restartAggressive`` when ``param_value`` is ``restart``; otherwise ``param_value`` unchanged.
+        """
+        if param_value == "restart":
+            return "restartAggressive"
+        return param_value
+
+    @staticmethod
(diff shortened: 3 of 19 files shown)


p3ck (commented):
Hi Mike,  I ran this pr through claude and it came back with the following.  this is the first time I'm doing this.  I looked at what it flagged and it all seems reasonable to me.

Thanks

---

Here's an initial review of PR #352. I was able to read the full `_cluster_settings.py` diff and the PR description/commit history directly. GitHub's authenticated-only rendering blocked the raw source for the individual module files, so some comments below are based on patterns visible in the diff view — flag anything that doesn't apply.

---

## Overall Assessment

This is a well-structured feature PR. The scope is appropriate (new functionality + refactor toward consistency), the description is clear, and the commit history tells a readable story. A few areas worth attention:

---

### ✅ What's Done Well

**`_cluster_settings.py` design is solid.** The `BaseVmOverrideChangeTracker` ABC is a clean abstraction. Using MoID as the dict key rather than VM name is the right call (names aren't unique). The `_is_desired_set_and_different_from_current` helper correctly handles the "don't overwrite if unset" idiom, which is easy to get wrong.

**`append` mode logic in `process_present_changes`** correctly separates "add to existing" vs "enforce exact list" behavior, which matches how Ansible modules typically express idempotency.

**Changelog fragment** is present and correctly categorized as `minor_changes`. ✅

**`galaxy.yml`** — adding `.ansible` to `build_ignore` is a sensible housekeeping addition.

---

### 🔍 Issues / Questions

**1. `format_override_specs_for_json` returns only `vm_moid` and `vm_name`**

The docstring says it returns `{"vm_moid": ..., "vm_name": ...}` — but this is used for the `changed` output in Ansible modules, right? If users want to know *what* changed about an override (not just which VM), stripping all the override attributes makes diff/audit harder. Consider whether the return value should include the full override details (restart priority, monitoring setting, etc.), at least optionally.

**2. `_GetMoId()` is a private pyVmomi method**

`change.key._GetMoId()` works today but `_GetMoId` is a private API. The public equivalent is typically `change.key._moId` or going through `vim.ManagedObject`. Worth a comment noting the tradeoff, or switching to the stable accessor if available in the minimum pyVmomi version (8.0.3.0.1 per releases).

**3. `process_absent_changes` doesn't populate `_final_override_moids` if no param VM is in `current`**

```python
self._final_override_moids = set(self.current_vm_overrides.keys())
for moid in self.param_vm_overrides.keys():
    if moid in self.current_vm_overrides:
        self.to_remove[moid] = ...
        self._final_override_moids.remove(moid)
```

This is correct, but `_final_override_moids` here represents "what remains after removals." If a caller later calls `process_present_changes` after `process_absent_changes`, they'd be operating on stale/conflicting state. Are these two methods meant to be called mutually exclusively (one for `state: absent`, one for `state: present`)? That constraint isn't enforced or documented. A comment or guard would help.

**4. No `__all__` export or module-level `__author__`/license header**

Other files in the collection likely have the standard GPL-3 header. This new utility module is missing it. Worth checking for consistency with the collection's standards.

**5. Commit hygiene — two commits named "test fixes"**

Minor: commits `6dde160` and `5c77a05` are both just called "test fixes." Fine for a feature branch but could be squashed before merge for a cleaner `git log`.

**6. `cluster_drs_vm_overrides` refactor — risk of regression**

The DRS modules are being refactored to align with the new HA pattern. This is the highest-risk part of the PR since it changes existing behavior. Make sure:
- The integration test in `vmware_cluster_drs_vm_overrides/tasks/main.yml` covers the same scenarios as before the refactor
- `check_mode` behavior is tested for both modules

---

### 📋 Questions for the Author

- Does `cluster_ha_vm_overrides` support all three HA per-VM settings: restart priority, host isolation response, and VM monitoring? The vSphere API's `ClusterDasVmConfigSpec` has all three — want to confirm all are exposed.
- Is there a `check_mode` path? The base class doesn't seem to define it — is that handled per-module?
- Is `append=True` the default state for `state: present`? If so, how does a user remove a single override without clearing all of them — is that what `state: absent` is for?

---

**Summary:** The core abstraction is well-designed and the approach of consolidating HA/DRS shared logic is the right call. The main things to address before merging are the return value completeness of `format_override_specs_for_json`, a note/fix around `_GetMoId()`, and ensuring the DRS refactor has regression coverage. Everything else is relatively minor.

mikemorency:
@p3ck Ive reviewed:
1. I think adding detailed change diffs would add a lot of complexity to the module. Since we are including `_info` modules, users could still get a before and after if they need it
2. Even AI knows more than Broadcom :) Despite being a private method, they actually recommend using it for the MOID
3. I added both a guard and a doc string to the methods mentioned here
4. I added the license header. The `__author__` and `__all__` attributes are not used in this collection currently
5. My commit hygiene is terrible, thank you AI for announcing it. I will squash when merging
6. The module was not previously released, so no risk of breaking workflows.

📋 Questions for the Author
I confirmed all of these questions are answered in the module docs

p3ck (approved):
Nice work!  Thanks Mike

Issue #234, opened 2025-09-02
Update the requirements to vcf-sdk


mariolenz:
I suggest to update the requirements to [vcf-sdk](https://pypi.org/project/vcf-sdk/). As far as I understand, this package includes pyvmomi, vmware-vcenter and vmware-vapi-common-client.

And maybe some additional stuff that might be useful in the future.

Additional information: I'm doing the same thing in `community.vmware`: ansible-collections/community.vmware#2457

@mikemorency Feel free to close this issue if you don't want to do this, and even the 3.0.0 milestone that I've created if you don't like it!

mikemorency:
Vcf requires python 3.9 or higher and ansible core 2.18 still supports py3.8 on remote hosts. So I think this collection cant switch until that is EOL?

mariolenz:
The modules of this collection don't run on remote (vCenter / ESXi) hosts but on the controller and connects to the remote host via API. Or am I wrong? So I don't see a problem there.

Anyway, I'm not suggesting to do this _now_. I suggest to do this in the next major release, that is 3.0.0. Maybe you're planning to drop ansible-core < 2.19 for 3.0.0, anyway, so we would be doubly sure there.

mikemorency:
The modules run on either the local (controller) host or a remote host, and then that host connects to vcenter/esxi. localhost is the more common use case IMO, but people can run these modules using a remote host due to security or network restrictions

mariolenz:
True, that's possible. But I don't know the requirements for remote hosts in this sense. I should think they are the same as for controllers, but I didn't find this documented anywhere.

I think this would be a breaking change. So we would have to wait for a new major release, anyway. But I'm sure that vcf-sdk is something worth considering. And so we don't forget, I've created a milestone for 3.0.0 and added this issue there.

When you start working on 3.0.0, we can review this again. And then we can implement this, decide against vcf-sdk and close the issue as wontfix, or move it to the 4.0.0 milestone. Whatever makes more sense then 😃

mariolenz:
@mikemorency I havn't watched the video yet, but it might be interesting: [VMware Cloud Foundation 9.1 SDK: Achieving Full-Stack Private Cloud and Fleet Automation](https://www.vmware.com/explore/video/6403821131112)

PR #353, opened 2026-05-13
chore: Remove setuptools < 82 pin; require vmware-vcenter >= 9.1.0.0

### LINKED ISSUE #350: _pkg_resources_shim.py: to work around setuptools issue
##### SUMMARY

Given we aren't getting any progress from Broadcom on the SDK being updated to support setuptools>83, do we want to consider patching the collection ourselves?

`plugins/module_utils/_pkg_resources_shim.py`
```python

  import sys
  import types
  from importlib.resources import files
  from importlib.metadata import version as _get_version, PackageNotFoundError

  def _ensure_pkg_resources():
      if 'pkg_resources' in sys.modules:
          return

      shim = types.ModuleType('pkg_resources')

      def resource_string(package, resource):
          return files(package).joinpath(resource).read_bytes()

      class Requirement:
          @staticmethod
          def parse(name):
              return name

      class _WorkingSet:
          def find(self, name):
              try:
                  ver = _get_version(name)
                  obj = types.SimpleNamespace(version=ver)
                  return obj
              except PackageNotFoundError:
                  return None

      shim.resource_string = resource_string
      shim.Requirement = Requirement
      shim.working_set = _WorkingSet()

      sys.modules['pkg_resources'] = shim


```

**Pros:**
  - Clean — vapi code doesn't know anything changed
  - Works for both crash sites (top-level and deferred imports)
  - No modifications to vendored code
  - Reversible — remove the shim when Broadcom ships the fix

**Cons:**
  - Faking an entire module in sys.modules is inherently hacky
  - If real pkg_resources is present (setuptools < 82), the shim is skipped — no conflict
  - Could mask issues if other code also uses pkg_resources and expects full API

**References**

* https://github.com/ansible-collections/community.vmware/issues/2548
* https://github.com/vmware/vcf-sdk-python/issues/6

@mariolenz Feel free to shoot down this ideas.

I'm suggesting it, as I'm facing other issues when trying to package Ansible downstream where I have more and more packages that need newer setuptools

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below

```

##### COLLECTION VERSION

```paste below

```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

gundalow:
##### SUMMARY

The setuptools < 82 pin (added in PR #325) worked around a pkg_resources import failure in vmware-vapi-runtime 2.61.2. setuptools 82.0.0 removed pkg_resources, which caused vmware-vapi-runtime to crash on import.

Pin vmware-vcenter >= 9.1.0.0 to ensure the fixed
vmware-vapi-runtime is always resolved; older versions of vmware-vcenter pull vmware-vapi-runtime <= 2.61.2 which would still break without the setuptools cap.

Fixes #350

Upstream references:
- vmware/vcf-sdk-python#6
- vmware/vsphere-automation-sdk-python#443

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
 pyVmomi>=8.0.3.0.1
-vmware-vcenter
-setuptools < 82
+vmware-vcenter>=9.1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
+# Versions >=9.1 are recommended, but not required
 pyVmomi>=8.0.3.0.1
-vmware-vcenter>=9.1.0.0
+vmware-vcenter
--- a/tests/integration/requirements.txt
+++ b/tests/integration/requirements.txt
@@ -2,5 +2,4 @@ ansible-core
 
 aiohttp
 pyVmomi>=8.0.3.0.1
-vmware-vcenter
-setuptools < 82
+vmware-vcenter>=9.1.0.0
--- a/tests/unit/requirements.txt
+++ b/tests/unit/requirements.txt
@@ -1,3 +1,2 @@
 pyVmomi>=8.0.3.0.1
-vmware-vcenter
-setuptools < 82
+vmware-vcenter>=9.1.0.0


p3ck (approved):
LGTM - Thanks

requirements.txt
```diff
@@ -1,3 +1,2 @@
 pyVmomi>=8.0.3.0.1
-vmware-vcenter
-setuptools < 82
+vmware-vcenter>=9.1.0.0
```
mariolenz: I think that this might be considered a breaking change. If people are stuck for whatever reason on Python 3.8 or 3.9, they'll run into a problem: [vmware-vcenter 8.0.3.0](https://pypi.org/project/vmware-vcenter/8.0.3.0/) supports them, but [vmware-vcenter 9.1.0.0](https://pypi.org/project/vmware-vcenter/9.1.0.0/) doesn't.

The integration and unit test requirements are OK, but personally I wouldn't do this in the collection requirements. It's basically a good idea, but in a minor version...? After all, some people might be happy (or are stuck with) running the collection on Python 3.8 / 3.9. So this would be breaking for them, which means a new major release. And just `vmware-vcenter` should, in most cases, install the latest version 9.1.0.0 anyway.

And if there are any changes to the requirements, I strongly suggest to use vcf-sdk: #234

Since this would also be a breaking change, we can require `vcf-sdk>9.1.0.0` then while we're at it. No problem.

mikemorency: Well then lets just unpin setup tools for now. We should do a major release soon-ish, but if unpinning setuptools is all that is required right now then it seems the best path forward

mikemorency: @mariolenz let me know what you think now, and we can merge

gundalow: Makes sense to me, thanks for the update.

mariolenz (approved):
LGTM

mariolenz:
@mikemorency @p3ck Do you have any plans to do a 2.9.0 release soon so people arent't stuck with setuptools<82 anymore?

mikemorency:
Im shooting for the end of the month

PR #356, opened 2026-05-19
feat: add key provider modules for vCenter

mikemorency:
#### SUMMARY
- Add `key_provider_native` to create, delete, and set default native key providers in vCenter (REST API for provider lifecycle; PyVmomi for default provider)
- Add `key_provider_standard` to manage standard key provider clusters and KMS servers (PyVmomi)
- Add `key_provider_info` to gather key provider details, with optional filtering by name or type
- Include unit tests and an integration test target (`vmware_key_provider`)

#### ISSUE TYPE
- New Module Pull Request

#### COMPONENT NAME
- plugins/modules/key_provider_native.py
- plugins/modules/key_provider_standard.py
- plugins/modules/key_provider_info.py
- tests/unit/plugins/modules/test_key_provider_native.py
- tests/unit/plugins/modules/test_key_provider_standard.py
- tests/unit/plugins/modules/test_key_provider_info.py
- tests/integration/targets/vmware_key_provider/
- changelogs/fragments/key-provider-modules.yml

#### ADDITIONAL INFORMATION
- Sanity tests pass on all three modules
- Unit tests cover present/absent, KMS server add/update/remove, default provider, and info filtering

> This PR body was generated using AI and reviewed by the author.

### DIFF
--- a/changelogs/fragments/key-provider-modules.yml
+++ b/changelogs/fragments/key-provider-modules.yml
@@ -1,0 +1,5 @@
+---
+minor_changes:
+  - key_provider_native - add module to manage native key providers in vCenter
+  - key_provider_standard - add module to manage standard key providers and KMS servers in vCenter
+  - key_provider_info - add module to gather information about key providers in vCenter
--- a/plugins/modules/key_provider_info.py
+++ b/plugins/modules/key_provider_info.py
@@ -1,0 +1,208 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2023, Ansible Cloud Team (@ansible-collections)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+---
+module: key_provider_info
+short_description: Gather information about one or more key providers in a vCenter instance.
+description:
+    - This module allows you to gather information about one or more key providers in a vCenter instance.
+
+author:
+    - Ansible Cloud Team (@ansible-collections)
+
+options:
+    provider_name:
+        description:
+            - The name or ID of the key provider to gather information about.
+            - If this is not provided, all key providers will be returned.
+            - Only one of O(provider_name) or O(type) is allowed.
+        type: str
+        required: false
+        aliases: [name, id]
+
+    type:
+        description:
+            - The type of key provider to gather information about.
+            - If this is not provided, all key provider types will be returned.
+            - Only one of O(provider_name) or O(type) is allowed.
+        type: str
+        required: false
+        choices: [standard, native]
+
+extends_documentation_fragment:
+    - vmware.vmware.base_options
+"""
+
+EXAMPLES = r"""
+- name: Gather information about all key providers
+  vmware.vmware.key_provider_info:
+    hostname: "{{ vcenter_hostname }}"
+    username: "{{ vcenter_username }}"
+    password: "{{ vcenter_password }}"
+
+- name: Gather information about a specific key provider
+  vmware.vmware.key_provider_info:
+    provider_name: my-standard-key-provider
+
+- name: Gather information about native key providers only
+  vmware.vmware.key_provider_info:
+    type: native
+
+- name: Gather information about standard key providers only
+  vmware.vmware.key_provider_info:
+    type: standard
+"""
+
+RETURN = r"""
+key_providers:
+    description:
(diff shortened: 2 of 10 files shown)


p3ck (commented):
---

## PR #356 Review — `feat: add key provider modules for vCenter`

### Overall Assessment

This is a well-scoped new-module PR adding `key_provider_native`, `key_provider_standard`, and `key_provider_info`. The PR description is clear, tests are included, and sanity passes. The mixed API approach (REST for native provider lifecycle, PyVmomi for setting default + standard providers) is the right call given vSphere API coverage.

One notable flag upfront: the first commit has a `Co-authored-by: Cursor <cursoragent@cursor.com>` tag, indicating significant AI authorship. The PR description acknowledges this too. That's fine as long as the code was thoroughly reviewed by the author — but it warrants extra scrutiny in review.

---

### ✅ What's Done Well

**`key_provider_info.py` is clean and well-structured.** The `TYPE_MAP` dict is a good pattern for translating internal vSphere type strings (`nativeProvider`, `vCenter`) to user-friendly values (`native`, `standard`). The `getattr(..., None)` null checks for optional fields like `tpmRequired` and `hasBackup` are correct.

**Mutually exclusive filters handled properly.** Using `mutually_exclusive=[["provider_name", "type"]]` in `AnsibleModule` is the right declarative approach rather than manual validation.

**`supports_check_mode=True`** on the info module is correct (it makes no changes).

**Changelog fragment** is present and well-formed. ✅

---

### 🔍 Issues / Questions

**1. `provider_name` aliases `id` — but filter only matches on ID**

In `gather_key_provider_info`, the filter is:
```python
if name_filter and kmip_cluster.clusterId.id != name_filter:
    continue
```
The docs say the parameter is "The name or ID of the key provider," but the code only matches on `clusterId.id`. For native providers managed via the REST API, the ID and name are typically the same. But for standard (KMIP) providers, it's worth confirming whether `clusterId.id` is always the human-readable name the user would know, or if there's a separate `name` field. If they can differ, the filter should try both.

**2. `backed_up` missing from the sample output for `my-native-key-provider`**

In the `RETURN` docstring sample, `my-native-key-provider` doesn't include `backed_up`, while `default-native-key-provider` does. The `kmip_cluster_to_dict` method conditionally includes `hasBackup` — so the sample is technically valid — but it's confusing because a user won't know which native providers will have it. The docs should note that `backed_up` is only present when the provider has been backed up at least once.

**3. Copyright year says `2023`**

All three modules have `# Copyright: (c) 2023, Ansible Cloud Team`. This work is from 2026. Minor, but worth correcting for accuracy.

**4. Mixed API strategy needs documentation**

`key_provider_native` uses the vCenter REST API for create/delete but drops into PyVmomi to set the default provider. This is a valid workaround (the REST API doesn't expose default provider management), but it should be noted in the module's `description` or `notes` section so users understand the dependency on both API paths and what permissions are required for each.

**5. `key_provider_standard` — no visibility into KMS certificate management**

Looking at how `community.vmware`'s `vcenter_standard_key_provider` handled this, standard KMS provider setup often requires uploading client certificates and trusting server certificates. If this module manages KMS server add/update/remove but doesn't handle certificate exchange, it should document that limitation prominently — users may be surprised that a "managed" KMS server isn't actually trusted until certs are exchanged out-of-band.

**6. Integration test target naming inconsistency**

The integration test target is named `vmware_key_provider` but the modules follow the new collection's naming convention without the `vmware_` prefix. Other targets in this collection should be checked — if they all use the `vmware_` prefix for targets, this is fine for consistency, but if some don't, it may be worth aligning.

---

### 📋 Questions for the Author

- For `key_provider_native`, what happens if you run with `state: present` and the provider already exists? Does it silently succeed (idempotent) or does the REST API return an error that needs to be handled?
- Is `tpm_required` settable at creation time only, or can it be updated on an existing native provider?
- Does the integration test clean up after itself (delete the test key providers), and does it test the `set_default` path?
- Are there any vSphere version constraints (e.g., native key providers require vSphere 7.0+) that should be documented in the module or in `meta/main.yml`?

---

**Summary:** Clean, useful addition. The main things to address are the `provider_name` filter potentially only matching on ID, the missing certificate management documentation for `key_provider_standard`, and the copyright year. The AI-assisted authorship makes thorough testing validation especially important here — confirm the integration tests cover the full lifecycle including idempotency and cleanup.

mikemorency:
@p3ck 

1. Good callout, the name and ID are the same thing. I updated the docs to clarify
2. Updated examples
3. Updated year
4. At the collection level, we say both are required. I added a REST requirement to the module though; it seems like that is what we did for other similar modules
5. vcenter should automatically try to trust the new KMS servers. I will add a note in the docs though
6. Nothing to do

Questions for the Author:
I validated these questions are answered in the docs. The exception is the one about testing the default provider path. I cant test that without potentially breaking the lab

p3ck (approved):

PR #357, opened 2026-05-19
feat(vm): add enable_vtpm parameter for virtual TPM management

mikemorency:
#### SUMMARY
- Add `enable_vtpm` parameter to the `vm` module to enable or disable virtual Trusted Platform Module (vTPM) on virtual machines.

#### ISSUE TYPE
- Feature Pull Request

#### COMPONENT NAME
- plugins/modules/vm.py
- plugins/module_utils/vm/objects/_vtpm.py
- plugins/module_utils/vm/parameter_handlers/device_linked/_vtpms.py

#### ADDITIONAL INFORMATION
- Unit tests added for the vTPM object and parameter handler.

---
*This PR description was drafted with AI assistance and reviewed by the author.*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,17 +53,10 @@ sanity: upgrade-collections
 
 .PHONY: units
 units: upgrade-collections
-<<<<<<< HEAD
 	cd $(COLLECTION_ROOT); \
 	ansible-test units --docker --python $(UNIT_PYTHON_VERSION) --coverage $(UNIT_TARGETS); \
 	ansible-test coverage combine --requirements --export tests/output/coverage/; \
 	ansible-test coverage report --requirements --docker --omit 'tests/*' --show-missing;
-=======
-	cd ~/.ansible/collections/ansible_collections/vmware/vmware; \
-	ansible-test units --docker --python 3.12 --coverage $(UNIT_TARGETS); \
-	ansible-test coverage combine --requirements --export tests/output/coverage/; \
-	ansible-test coverage report --requirements --docker --omit 'tests/*' --show-missing
->>>>>>> c9d42de (add vtpm support to vm module)
 
 .PHONY: units-coverage
 units-coverage: units
--- a/Makefile
+++ b/Makefile
@@ -53,10 +53,17 @@ sanity: upgrade-collections
 
 .PHONY: units
 units: upgrade-collections
+<<<<<<< HEAD
 	cd $(COLLECTION_ROOT); \
 	ansible-test units --docker --python $(UNIT_PYTHON_VERSION) --coverage $(UNIT_TARGETS); \
 	ansible-test coverage combine --requirements --export tests/output/coverage/; \
 	ansible-test coverage report --requirements --docker --omit 'tests/*' --show-missing;
+=======
+	cd ~/.ansible/collections/ansible_collections/vmware/vmware; \
+	ansible-test units --docker --python 3.12 --coverage $(UNIT_TARGETS); \
+	ansible-test coverage combine --requirements --export tests/output/coverage/; \
+	ansible-test coverage report --requirements --docker --omit 'tests/*' --show-missing
+>>>>>>> c9d42de (add vtpm support to vm module)
 
 .PHONY: units-coverage
 units-coverage: units
--- a/changelogs/fragments/05192026-vm-vtpm.yml
+++ b/changelogs/fragments/05192026-vm-vtpm.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - vm - Add enable_vtpm parameter to enable or disable virtual Trusted Platform Module (vTPM) on virtual machines.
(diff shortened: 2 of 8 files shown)


p3ck (commented):
## PR #357 Review — `feat(vm): add enable_vtpm parameter for virtual TPM management`

### Overall Assessment

This is a clean, focused feature addition that fits well into the existing `vm` module architecture. The separation between the `Vtpm` object (`_vtpm.py`) and the `VtpmParameterHandler` (`_vtpms.py`) follows the established pattern for device-linked handlers, the constraints are well-reasoned, and the integration test coverage is solid. This is one of the more straightforward PRs of the three.

---

### ✅ What's Done Well

**Correct constraint checking.** Validating both EFI firmware and snapshot presence before enabling vTPM reflects the actual vSphere requirements. Importantly, the firmware check gracefully falls back to the live VM's `config.firmware` when `boot_firmware` isn't specified in the params — that's the right behavior.

**`to_update_spec` returns `None` correctly.** Since vTPM has no mutable attributes, returning `None` from `to_update_spec` and documenting why is the right approach. The `differs_from_live_object` method that always returns `False` when a device is linked is logically consistent with this.

**`link_vm_device` duplicate detection** is a nice defensive guard. If vSphere somehow ends up with two vTPM devices (it shouldn't), the module will fail clearly rather than silently misbehave.

**Documentation in `vm.py` is thorough.** The `enable_vtpm` param docs mention EFI requirement, key provider prerequisite, snapshot constraint, and power cycling requirement. That's everything a user needs to know upfront.

**Integration tests cover the full round-trip** — add vTPM (checking `objects_to_add[0].object_type == 'vtpm'`), then remove it (checking `objects_to_remove[0].object_type == 'vtpm'`). The idempotency assertions are present too.

**Makefile fix** for `--requirements` on coverage is a good catch unrelated to the main feature.

---

### 🔍 Issues / Questions

**1. SonarCloud warnings are worth addressing before merge (3 flagged)**

SonarCloud flagged 4 new issues, all minor but real:

- Line 74 (`_vtpms.py`): `dict()` constructor — should be `{}`. Change `self.params.get("vm_options", dict())` to `self.params.get("vm_options", {})`. Same issue flagged on `vm.py` line 1357 for `dict(type='bool', required=False)` → this one is actually standard Ansible module arg spec style and shouldn't be changed, so it may be a false positive. The `dict()` in the get-default case is the real one to fix.
- Lines 92 and 123 (`_vtpms.py`): redundant bare `return` at end of methods. Both are at the natural end of a function body and can simply be removed.

These are all trivially fixable and would clean up the SonarCloud report.

**2. `verify_parameter_constraints` doesn't check firmware for new VM creation**

The constraint check has this guard:
```python
if self.vm is not None:
    if boot_firmware is None:
        boot_firmware = self.vm.config.firmware
```

For a new VM (`self.vm is None`), if the user passes `enable_vtpm: true` but doesn't set `boot_firmware: efi`, `boot_firmware` stays `None` and the EFI check below will evaluate `None != "efi"` as `True`, triggering a `fail_json`. That's actually the correct outcome — but the error message will say "vTPM requires EFI boot firmware" with `details: {"firmware": None}`, which is potentially confusing. Consider a more specific message for the new-VM case: "vTPM requires EFI boot firmware. Set `vm_options.boot_firmware: efi`."

**3. Integration test only runs in the existing `basic_vm_update_tests.yml` — no isolated vTPM test**

The vTPM test is embedded in the broader update flow, which means the vTPM state depends on whatever the VM state is at that point in the playbook. It would be more resilient as a dedicated test block or file (`vtpm_tests.yml`) so it can be run and debugged independently, and so failures are clearly attributable to vTPM logic specifically.

**4. No test for the snapshot constraint**

The `verify_parameter_constraints` snapshot check is only tested at the unit level (presumably — unit tests weren't fully visible in the diff). The integration test doesn't appear to test the error case where `enable_vtpm: true` is attempted on a VM with snapshots. Not a blocker, but worth noting.

**5. Changelog fragment naming**

The fragment is named `05192026-vm-vtpm.yml` (date-prefixed). Most others in this repo use ticket/feature names. Either convention works, but it's inconsistent — worth aligning with the project convention if there is one.

---

### 📋 Minor Notes

The `_to_module_output` for `Vtpm` only returns `object_type` and `label`. For consistency with other device objects, it might be worth confirming there's nothing else useful to return (like whether the device was linked or newly added), though for vTPM specifically there's really nothing else meaningful to expose.

---

**Summary:** Very solid PR. The three SonarCloud issues (redundant `return`s and `dict()` literal) should be fixed since they're trivial, and the firmware error message for new VM creation could be made more actionable. Everything else is clean and the integration test coverage is good. This is ready to merge after addressing those minor items.

mikemorency:
@p3ck 
I addressed the sonarqube findings. The other suggestions are either not applicable or covered by the module documentation

p3ck (approved):
LGTM!

PR #362, opened 2026-05-28
vm_snapshot: Add timeout parameter

mariolenz:
##### SUMMARY
Fixes #361

Add a timeout for snapshot operations to override the default 1 hour (3600 seconds) the module waits for the task to succeed before failing.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
vm_snapshot

##### ADDITIONAL INFORMATION
I didn't find the time to test this, but since the change is very simple and straightforward I think this isn't really necessary.

PR #363, opened 2026-05-28
CI: Use actions/checkout@v6

mariolenz:
I see  this warning in the CI:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Let's try to update to a newer version of `actions/checkout`

##### ADDITIONAL INFORMATION
https://github.com/ansible-collections/vmware.vmware/pull/362#issuecomment-4566159081

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -21,7 +21,7 @@ jobs:
           sudo apt-get install podman
 
       - name: Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Run
         run: |
--- a/.github/workflows/eco-vcenter-ci.yaml
+++ b/.github/workflows/eco-vcenter-ci.yaml
@@ -22,7 +22,7 @@ jobs:
       groups: ${{ steps.set-matrix.outputs.groups }}
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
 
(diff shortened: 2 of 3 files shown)


mikemorency:
i think the eco-vcenter-ci workflow wont pick these changes up until they are on main. But the other workflows should, and if they pass without the warnings then I feel good about merging

mariolenz:
I'm not really sure about the eco-vcenter-ci workflow, either. But if the rest of the CI is happy, we can merge and somehow trigger eco-vcenter-ci to see if it still works, too. If yes, everything's OK. If not, we'll just try again with v5.

If it's not working with v5 or v6, we have to have a closer look. As far as I understand, v4 won't work in a couple of months anymore.

mikemorency (approved):

mariolenz:
FYI [this run](https://github.com/ansible-collections/vmware.vmware/actions/runs/26590038113) still showed the warnings about Node.js 20 actions beeing deprecated, but [the last one](https://github.com/ansible-collections/vmware.vmware/actions/runs/26599114287) doesn't. And since I don't see any failures there, the CI seems to be happy with `actions/checkout@v6`.

PR #412, opened 2026-09-08
Added missing modules to action_groups

AllRWeak:
##### SUMMARY
This PR adds the new `vm_snapshot_info` and other missing modules to the runtime.yml so [Ansible module defaults](https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_module_defaults.html) would work for them. 

Related to `community.vmware.vmware_guest_snapshot_info` being deprecated but the new modules in  `vmware.vmware` collection not being added to action_groups

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`vm_snapshot_info` (and other missing)

##### ADDITIONAL INFORMATION
Ansible has a feature to define module defaults like this for the whole playbook/project/etc. so these keys don't need to be added to each module used.
```
vmware_defaults:
  hostname:
  username:
  password:
  ```
  
Currently many modules were missing meta/runtime.yml list for the module defaults to work. This PR adds them there. Also sorted them alphabetically.


mikemorency (approved):
thanks, ive been meaning to do this and keep forgetting

PR #413, opened 2026-09-09
Add check mode support to the vm module

mikemorency:
##### SUMMARY
The `vm` module previously ignored Ansible's check mode and would attempt to create, update, or delete a VM even during a check-mode run. This change makes the module respect check mode so that create, update, and delete operations report whether a change *would* occur without actually modifying the VM.

- `_run_task` now short-circuits and returns an empty result when running in check mode, so no vSphere task is executed.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vm

##### ADDITIONAL INFORMATION
Adds unit tests (`tests/unit/plugins/modules/test_vm.py`) and integration tests (`tests/integration/targets/vm/tasks/check_mode_tests.yml`) covering check mode for create, update, and delete operations.


plugins/modules/vm.py
```diff
@@ -1171,6 +1172,10 @@ def _try_to_run_task(self, task_func, action, task_kwargs=None):
         Returns:
             The vSphere task result.
         """
+        if self.module.check_mode:
+            # dont run the task, just return a fake result
+            return {}
```
p3ck, line 1177: Couldn't you return {'result': None} and then you wouldn't need the condition on line 1064?

mikemorency: if its an update operation, that would set the vm outputs to None instead of the existing vm

PR #414, opened 2026-09-12
Fix node identity in extensions/audit/event_query.yml

fistilis:
##### SUMMARY

Nothing in this repository executes the jq in `extensions/audit/event_query.yml`. It runs inside the Automation Platform controller ([`awx/main/tasks/host_indirect.py`](https://github.com/ansible/awx/blob/devel/awx/main/tasks/host_indirect.py)) and its output feeds subscription node counting, so a query can be wrong for a long time without anything failing.

I ran every query in the file against the shape its module really returns, taking the payloads from each module's `exit_json()` and `RETURN` block. **13 of the 38 entries reported no node at all, and one virtual machine produced 7 audit rows.**

```
before                          after
  COUNTED      25                 COUNTED      32
  DROPPED      10
  EVENT ERROR   3

ONE virtual machine (vm-79828) touched by every VM module
  -> 7 audit rows                 -> 1 audit row
```

##### THE THREE CONSTRAINTS BEING VIOLATED

The controller imposes these, and violating any of them produces no error:

1. `canonical_facts` **cannot contain null at any depth.** `get_hashable_form()` accepts `int/float/str/bool/dict/list/tuple` and raises `UnhashableFacts` on anything else, including `null`; [the caller catches it and `continue`s](https://github.com/ansible/awx/blob/devel/awx/main/tasks/host_indirect.py). One null in one field discards the whole node — logged once per job at `INFO`, job still green.
2. **A jq error is caught at the *event* level**, not per module. So one query that indexes a bool or a string discards every *other* module's audit records for that event too.
3. `canonical_facts` **is the only dedup key** (`results[hashable_facts]`); `facts` and `name` are taken from the first record seen. Anything in it that is not identity is an overcount.

##### 1. Wrong source key, raising a jq error (3 modules, and they take the event with them)

| module | reads | actual type | error |
|---|---|---|---|
| `vm` | `.power_cycled_for_update` | `bool`, always returned | `Cannot index boolean with string "moid"` |
| `license_info` | `.licenses[]` | list of license key **strings** (`VcenterLicenseMgr.list_keys` returns `item.licenseKey`) | `Cannot index string with string "id"` |
| `tag_associations` | `.added_tags[]` | list of tag id **strings** | `Cannot index string with string "id"` |

`vm` is the collection's primary VM lifecycle module. The identity its query wants has always been under `.vm` (`{moid, name}`). For `tag_associations` the node is under `.object`, whose moid prefix also tells you what kind of object was tagged.

##### 2. Wrong source key, yielding null (7 modules silently dropped)

| module | reads | where the identifier actually is |
|---|---|---|
| `esxi_maintenance_mode` | `.result` — maintenance-mode *task* info | `.host.moid` |
| `import_content_library_ovf` | `.` — the whole result | `.library_item.id` |
| `import_content_library_iso` | `.` | `.library_item.id` |
| `cluster_info` | `.clusters.moid` | `clusters` is a dict keyed by cluster **name** |
| `tags` | `.vsphere_tags.id` | dict keyed by tag id |
| `tag_categories` | `.vsphere_tag_categories.id` | dict keyed by category id |
| `vm_list_group_by_clusters_info` | `.id` | `{cluster: {folder: [vm]}}`, and the id key is `vm` |

On that last one: the entries are `{k: str(v) for k, v in vars(VM.Summary)}`, whose id field is named `vm`. The module's `RETURN` sample documents `moid` instead, which is presumably where the query got it — the query now accepts `(.vm // .moid)` so it works either way. With `detailed_vms=true` the entries are `VM.Info`, which carries no id at all; those runs now report nothing, which is correct, since there is no identity to deduplicate on.

##### 3. The module name was in the dedup key

Every entry had `canonical_facts: {module: "<name>", moid: ...}`. Since `canonical_facts` is the sole dedup key, one VM touched by `vm`, `vm_powerstate`, `vm_snapshot`, `vm_advanced_settings`, `vm_apply_customization`, `deploy_content_library_ovf` and `deploy_folder_template` became **7 nodes**. `module` moved to `facts`, which is not hashed.

##### 4. Taxonomy

`deploy_content_library_ovf`, `deploy_content_library_template` and `deploy_folder_template` all return `vm: {moid, name}` — their own `RETURN` blocks say "Identifying information about the vm" — but were labelled `device_type: content_library` / `folder` (and `infra_bucket: governance_ops` for the last). Now `compute` / `virtual_machine`.

##### 5. Removals — please push back if you disagree

I removed six entries whose modules return no identifier that can distinguish one node from another, so no query over them could have worked. Each is documented in a comment block at the bottom of the file and is a one-line revert:

- `folder_template_from_vm` — `result = dict(changed=...)` and nothing else.
- `license_info` — license key strings; also the event-level jq error above. A license is not a node.
- `vm_portgroup_info` — returns `{vm_name: [portgroup]}`. The only identity is the mutable VM *name*, which is not the moid every other query uses, so counting it would add a second row for a VM already counted correctly.
- `vcsa_settings` — DNS/NTP settings, no identifier. The appliance itself is reported by `appliance_info`.
- `vcsa_backup_schedule`, `vcsa_backup_schedule_info` — the schedule id is `"default"` on every appliance ("*Your VCSA may only allow for one schedule... its name is 'default'*"), so `canonical_facts` was byte-identical across every vCenter in a fleet and collapsed all of them into a single row.

##### 6. Defensive changes applied to every remaining entry

- `select(type == "object")` before indexing, so a source key that is not the expected shape drops cleanly instead of killing the event for every module.
- `select(<id> != null)`, so a record with no identity is dropped explicitly rather than emitting `canonical_facts: {moid: null}`.
- `canonical_facts.moid` for vSphere managed object ids and `canonical_facts.id` for Automation API uuids, used consistently — so the same VM reached through any module deduplicates to one row.
- `name: (.name // (<id> | tostring))`, which is never null.

##### 7. The test

`tests/unit/audit/test_event_query_contract.py` asserts the controller's constraints statically — name non-null, `canonical_facts` present and null-free at any depth, `canonical_facts` holds identity only, taxonomy keys present and `lowercase_with_underscores`. It is stdlib + pytest + PyYAML, needs no credentials and no vCenter, and runs in 0.2s (225 assertions).

A snapshot test would not have caught any of this: capturing expected output *from actual output* makes the assertion agree with the output by construction, so it detects drift but never wrongness. Every defect above was present on day one.

Reverting any single fix in this PR makes it fail:

```
$ # put the module name back into canonical_facts
FAILED test_canonical_facts_holds_identity_only[vmware.vmware.vm_snapshot]
$ # put `(.moid // null)` back, unguarded
FAILED test_no_field_in_canonical_facts_can_be_null[vmware.vmware.esxi_host]
$ # drop device_type / de-normalize infra_bucket
FAILED test_facts_carry_the_full_taxonomy[vmware.vmware.cluster_dpm]
FAILED test_taxonomy_values_are_normalized[vmware.vmware.folder]
```

Tracked internally as [AAP-92639](https://issues.redhat.com/browse/AAP-92639); part of a cross-collection audit of every collection that reports indirect nodes ([AAP-92620](https://issues.redhat.com/browse/AAP-92620)).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
event_query


mikemorency:
Unit tests should mimic the same folder path as what they are testing; these should be in the extensions folder. I also think. a filename of test_event_query.py would suffice

mikemorency (changes requested):
please migrate unit tests to the correct location and address ci failures

fistilis:
Thanks @mikemorency — both addressed in 7d88917.

**Test location:** moved `tests/unit/audit/` → `tests/unit/extensions/audit/`, so it mirrors the source path of `extensions/audit/event_query.yml` the same way `tests/unit/extensions/eda/` mirrors `extensions/eda/`. `COLLECTION_ROOT` adjusted for the extra level; the 225 checks still pass locally.

**CI:** the sanity failure was my changelog fragment, not the query change. One bugfix entry contained `"<key>": null` in an unquoted scalar, so YAML parsed the list item as a mapping and the changelog test rejected it (`section "bugfixes" list items must be type str not dict`). Reworded. Also fixed a pylint `disallowed-name` on a throwaway `_` in the test.

mikemorency (approved):

PR #416, opened 2026-09-14
inventory/vms: add gather_path option and bulk tag fetching

scouture-bw:
## Summary

Two independent serial-per-VM vCenter API patterns dominate `ansible-inventory` runtime when caching is disabled. On a 30-VM lab vCenter over a local connection this PR reduces runtime from ~50 s to ~5 s with `gather_tags: true`.

### 1. Folder path traversal via pyVmomi lazy loading

`get_folder_path_of_vsphere_object` traverses each VM's parent folder chain by accessing `.name` and `.parent` on pyVmomi managed object references. pyVmomi lazy-loads these properties on demand, so each attribute access issues a synchronous vCenter RPC. With ~4 folder levels per VM this produces 4 blocking round-trips per VM, scaling linearly with inventory size.

### 2. Per-VM REST tag fetching

`populate_from_vcenter` called `add_tags_to_object_properties` (and thus `list_attached_tags`) once per VM, then resolved each tag object individually — O(n × tags) REST round-trips.

### `gather_path` option (`plugins/inventory/vms.py`, `plugins/inventory_utils/_base.py`)

- New `gather_path` bool option, **default `true`** — fully backward-compatible.
- When `false`, `create_from_vcenter_object` skips `get_folder_path_of_vsphere_object` entirely and sets `path` to an empty string.
- Safe to set `false` when `path` is not referenced in `filter_expressions`, `compose`, `keyed_groups`, `hostnames`, `groups`, or `group_by_paths`.

### Bulk tag fetching (`plugins/module_utils/clients/rest.py`, `plugins/inventory_utils/_base.py`, `plugins/inventory/vms.py`)

- `VmwareRestClient.get_tags_for_vm_moids_bulk`: calls `list_attached_tags_on_objects` once for all VMs, then resolves each unique tag object exactly once. Returns a `{moid: [tag_obj, ...]}` mapping.
- `VmwareInventoryBase.add_tags_to_objects_bulk`: distributes bulk results across host objects, resolving category names once per unique category.
- `populate_from_vcenter` now collects all VM objects first, then calls the bulk tag method in a single pass — replacing the per-VM `add_tags_to_object_properties` call inside the loop.
- The existing `add_tags_to_object_properties` is retained unchanged for backward compatibility.

## Benchmark

Measured with `ansible-inventory -i <config> --graph` against production vCenters over varying network distances (`gather_tags: true`, `gather_path: false`):

| Site | Runtime |
|---|---|
| Local / US East | 3.7 – 5.4 s |
| US West / Canada | 7.0 – 8.1 s |
| Europe | 7.0 – 7.1 s |
| Asia-Pacific | 19.8 – 21.0 s |

Pre-patch baseline: ~50 s locally, 2–3 min on higher-latency connections.

## Testing

Unit tests added/updated across all three modified files:

**`tests/unit/plugins/module_utils/clients/test_rest.py`**
- `test_get_tags_for_vm_moids_bulk_empty_input` — empty input returns `{}` without calling `list_attached_tags_on_objects`
- `test_get_tags_for_vm_moids_bulk` — verifies `list_attached_tags_on_objects` is called once and `tag_service.get` is called once per unique tag id (not once per vm×tag pair)
- `test_get_tags_for_vm_moids_bulk_vm_with_no_tags` — VMs absent from the bulk response receive an empty tag list

**`tests/unit/plugins/inventory_utils/test_base.py`**
- `test_create_from_vcenter_object_skips_path_when_gather_path_false` — verifies `get_folder_path_of_vsphere_object` is not called and `path` is `''` when `gather_path=False`
- `test_add_tags_to_objects_bulk` — verifies `tags` and `tags_by_category` are populated correctly across multiple hosts with different tag combinations
- `test_add_tags_to_objects_bulk_caches_category_names` — verifies `tag_category_service.get` is called only once per unique category, even when multiple VMs share the same tag

**`tests/unit/plugins/inventory/test_vms.py`**
- Updated `test_populate_from_vcenter_with_gather_tags` — fixed stale assertion from `add_tags_to_object_properties` (per-VM) to `add_tags_to_objects_bulk([vm])` (batch)
- `test_populate_from_vcenter_with_gather_path_false` — verifies `create_from_vcenter_object` is called with `gather_path=False`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


plugins/inventory_utils/_base.py
```diff
@@ -60,16 +60,20 @@ def create_from_cache(cls, inventory_hostname, properties):
...
         Otherwise properties are read from the object via vmware_obj_to_json().
+
+        gather_path controls whether the vSphere folder path is computed. Path traversal
+        uses pyVmomi lazy loading (one RPC per folder level per VM), so set it to False
+        when the path variable is not needed.
         """
         host = cls()
         host.object = vmware_object
-        host.path = get_folder_path_of_vsphere_object(vmware_object)
+        host.path = get_folder_path_of_vsphere_object(vmware_object) if gather_path else ''
```
mikemorency, line 76: this attribute should be removed if we dont collect it, not set to an empty string

plugins/inventory_utils/_base.py
```diff
@@ -60,16 +60,20 @@ def create_from_cache(cls, inventory_hostname, properties):
         return host
 
     @classmethod
-    def create_from_vcenter_object(cls, vmware_object, properties_to_gather, pyvmomi_client, prop_set=None):
+    def create_from_vcenter_object(cls, vmware_object, properties_to_gather, pyvmomi_client, prop_set=None, gather_path=True):
```
mikemorency, line 63: Its cleaner to avoid using an input parameter like this instead of the `options` method. You will also need to implement this option in the esxi host inventory

plugins/inventory/vms.py
```diff
@@ -346,10 +369,11 @@ def populate_from_vcenter(self):
         """
         hostvars = {}
         properties_to_gather = self.parse_properties_param()
+        gather_path = self.get_option("gather_path")
         self.initialize_pyvmomi_client()
-        if self.get_option("gather_tags"):
-            self.initialize_rest_client()
```
mikemorency, line 351: i dont think this should be moved

plugins/inventory/vms.py
```diff
@@ -359,11 +383,15 @@ def populate_from_vcenter(self):
...
             )
+            vms.append(vm)
 
-            if self.get_option("gather_tags"):
-                self.add_tags_to_object_properties(vm)
+        if self.get_option("gather_tags"):
+            self.initialize_rest_client()
+            self.add_tags_to_objects_bulk(vms)
 
+        for vm in vms:
```
mikemorency, line 394: just move this into the original loop on line 377. theres no reason to loop again.

mikemorency (changes requested):
I think these would be better as separate changes. Heres some general feedback:

- The `gather_path` option seems like a minimal performance improvement but also a smaller scoped change. I left minor feedback
- The bulk tags method is interesting but definitely larger scope. The current implementation seems a bit hard to follow. There are many loops, comprehensions, if/else that add complexity. If you feel this is outside of your wheelhouse, I am happy to help
- Both improvements should be added to both inventory plugins

scouture-bw:
Splitting this into two focused PRs based on your review feedback, @mikemorency:

- **gather_path option:** #417
- **Bulk tag fetching:** #418

Closing this in favour of those two. Thanks for the detailed review!

PR #417, opened 2026-09-15
inventory: add gather_path option to skip folder path computation

scouture-bw:
Adds a `gather_path` option (default `true`) to both the `vms` and `esxi_hosts` inventory plugins.

When set to `false`, `get_folder_path_of_vsphere_object` is not called and the `path` key is omitted from host properties entirely. Path computation traverses the parent folder chain via pyVmomi lazy loading — one synchronous vCenter RPC per folder level per object — so disabling it saves roughly 1–2 seconds per object on a local network connection when `path` is not referenced in any of `filter_expressions`, `compose`, `keyed_groups`, `hostnames`, `groups`, or `group_by_paths`.

The option is handled in `VmwareInventoryHost.create_from_vcenter_object` in `_base.py` (via an optional `gather_path=True` kwarg), keeping the logic close to where `path` is actually set.

Addresses feedback from #416. Splitting that PR into this one (gather_path) and a separate bulk tag-fetching PR.


plugins/inventory/esxi_hosts.py
```diff
@@ -273,6 +285,9 @@ def populate_from_vcenter(self):
             if self._host_connection_state(esxi_host) in ("disconnected", "notResponding"):
                 continue
 
+            if not gather_path:
+                esxi_host.properties.pop('path', None)
```
mikemorency: the reasoning and impact of this is confusing since it is distanced from where the path attribute is actually set. In your original pr you modified this in the inventory utils base class; it makes more sense to do this there

mikemorency (changes requested):

plugins/inventory/esxi_hosts.py
```diff
@@ -30,6 +30,17 @@
...
+        description:
+            - If true, the vSphere folder path of each ESXi host is computed and stored in the C(path)
+              host variable. This requires traversing the parent folder chain via pyVmomi lazy
+              loading, which makes one network round-trip per folder level per host and dominates
+              inventory run time when caching is disabled.
+            - Set to false when C(path) is not referenced in C(filter_expressions), C(compose),
+              C(keyed_groups), C(hostnames), C(groups), or C(group_by_paths). Saves roughly
+              1-2 seconds per host on a local network connection.
+        default: true
+        type: bool
```
mikemorency, line 40:
```suggestion
    gather_path:
        description:
            - If true, the vSphere folder path of each ESXi host is computed and stored in the C(path)
              host variable.
            - If false, the C(path) attribute is not evaluated, potentially saving execution time and bandwidth.
            - This is required if C(path) referenced elsewhere in the inventory configuration, or O(group_by_paths) is C(True)
        default: true
        type: bool
```

plugins/inventory/vms.py
```diff
@@ -54,6 +54,17 @@
...
+        description:
+            - If true, the vSphere folder path of each VM is computed and stored in the C(path)
+              host variable. This requires traversing the parent folder chain via pyVmomi lazy
+              loading, which makes one network round-trip per folder level per VM and dominates
+              inventory run time when caching is disabled.
+            - Set to false when C(path) is not referenced in C(filter_expressions), C(compose),
+              C(keyed_groups), C(hostnames), C(groups), or C(group_by_paths). Saves roughly
+              1-2 seconds per VM on a local network connection.
+        default: true
+        type: bool
```
mikemorency, line 64:
```suggestion
    gather_path:
        description:
            - If true, the vSphere folder path of each VM is computed and stored in the C(path)
              host variable.
            - If false, the C(path) attribute is not evaluated, potentially saving execution time and bandwidth.
            - This is required if C(path) referenced elsewhere in the inventory configuration, or O(group_by_paths) is C(True)
        default: true
        type: bool
```

plugins/inventory_utils/_base.py
```diff
@@ -69,7 +69,8 @@ def create_from_vcenter_object(cls, vmware_object, properties_to_gather, pyvmomi
         """
         host = cls()
         host.object = vmware_object
-        host.path = get_folder_path_of_vsphere_object(vmware_object)
+        if gather_path:
+            host.path = get_folder_path_of_vsphere_object(vmware_object)
```
mikemorency, line 73: Back to the original feedback :)
`gather_path` is not needed as a method parameter. You should be able to call `host.get_option("gather_path")` instead

mikemorency (commented):
I think other than my new comment, this one is also good to go

plugins/inventory_utils/_base.py
scouture-bw: Hi Mike, thanks for the follow-up! Just to make sure I understand — `VmwareInventoryHost` (the host data class) doesn't inherit from `BaseInventoryPlugin` and doesn't have a `get_option` method, so calling `host.get_option("gather_path")` inside `create_from_vcenter_object` would raise an `AttributeError`.

Would the following work for you: remove `gather_path` from `create_from_vcenter_object`, and instead add a small helper method to `VmwareInventoryBase` (e.g. `_apply_gather_path(self, host)`) that calls `self.get_option("gather_path")` and pops `path` from the host properties when it's `False`? Both plugins would call it right after `create_from_vcenter_object`. The logic would stay in `_base.py` close to where `path` is set, and the classmethod stays parameter-free — the only trade-off is that the pyVmomi folder traversal still runs even when `gather_path=False` (we'd be cleaning up after the fact rather than skipping the computation).

Or if you had a different design in mind, happy to hear it!

mikemorency: my mistake :) lets go with what you currently have

tests/unit/plugins/inventory/test_vms.py
```diff
@@ -164,3 +164,27 @@ def test_populate_from_vcenter_with_gather_compute_objects(self, mocker):
...
+        )
+        vm = VmInventoryHost()
+        mocker.patch.object(VmInventoryHost, 'create_from_vcenter_object', return_value=vm)
+        mocker.patch.object(inventory_module, 'set_inventory_hostname')
+        mocker.patch.object(inventory_module, 'add_host_object_from_vcenter_to_inventory')
+        inventory_module.get_option = mocker.Mock(return_value=False)
+
+        inventory_module.populate_from_vcenter()
+
+        _, call_kwargs = VmInventoryHost.create_from_vcenter_object.call_args
```
mikemorency:
```suggestion
        _, call_kwargs = VmInventoryHost.create_from_vcenter_object.call_args  # pylint: disable=disallowed-name
```

tests/unit/plugins/inventory/test_esxi_hosts.py
```diff
@@ -188,3 +188,31 @@ def test_populate_from_vcenter_with_gather_tags(self, mocker):
...
+            return_value=EsxiInventoryHost(),
+        )
+        mocker.patch.object(inventory_module, '_host_connection_state', return_value='connected')
+        mocker.patch.object(inventory_module, 'set_inventory_hostname')
+        mocker.patch.object(inventory_module, 'add_host_object_from_vcenter_to_inventory')
+        inventory_module.get_option = mocker.Mock(return_value=False)
+
+        inventory_module.populate_from_vcenter()
+
+        _, call_kwargs = EsxiInventoryHost.create_from_vcenter_object.call_args
```
mikemorency:
```suggestion
        _, call_kwargs = EsxiInventoryHost.create_from_vcenter_object.call_args  # pylint: disable=disallowed-name
```

mikemorency (approved):

